PR #1, opened 2022-07-14
Feature/init phx project

pkrucz00:
Created a phoenix project and "merged" it with the membrane template. 
Additionally created a `docker-compose` file for quicker start of the postgres database. To start the app do:
```bash
docker-compose up
mix ecto.create
mix phx.server
```


Rados13:
On CircleCI linter and tests fails. 
- Fix linter issues. There are good descriptions of what is wrong.
- For tests, the problem is that it can't connect to the database. I see two possible solutions. The first is to use SQLite for the test environment, second is to change the CI config so that it will run the docker image during tests. I think that we will not use the advanced features of Postgres so it is up to you, which option will you choose.

pkrucz00:
@Rados13 I don't see the merit of changing test environment db to mysql. It's possible but nonetheless adding dockerized mysql db to circleci is still needed.

Rados13 (approved):

PR #3, opened 2022-07-15
Live 28 add form for creating and updating a webinar event

Karolk99:


assets/js/Form.tsx
```diff
@@ -0,0 +1,81 @@
...
+        "Content-Type": "application/json",
+        'Accept': 'application/json'
+        },
+      body: JSON.stringify(eventInfo)
+    }).then((response) => {
+            return response.json();
+    }).then((data) => {
+      setLinks(JSON.parse(data));
+    });
+  };
```
Rados13, line 45: I think that we should make some validation, whether all required fields are set (in our case it will be only title and date), before sending the request.

assets/js/Form.tsx
```diff
@@ -0,0 +1,81 @@
...
+            placeholder='Enter participant'
+            value={currParticipant}
+            onChange={handleInputParticipant}
+          />
+          <InputRightElement width='4.5rem'>
+            <Button h='1.75rem' size='sm' onClick={handleAddButton}>
+              Add
+            </Button>
+          </InputRightElement>
+        </InputGroup>
```
Rados13, line 72: It would be nice to have the possibility to see what participants/presenters we already added.

Rados13 (changes requested):

assets/js/Form.tsx
```diff
@@ -0,0 +1,103 @@
...
+        <UnorderedList>
+        {eventInfo.presenters &&
+          eventInfo.presenters.map((presenter, idx) => <ListItem key={idx}>{presenter}</ListItem>)}
+        </UnorderedList>
```
Rados13, line 94: I think that it would be nice to have the possibility to remove a participant because someone can make a typo.

assets/js/app.tsx
```diff
@@ -1,12 +1,14 @@
...
+import { ChakraProvider, Center } from "@chakra-ui/react";
+import Form from "./Form";
 
 ReactDOM.createRoot(document.getElementById('main')!).render(
   <React.StrictMode>
     <ChakraProvider>
-      <Hello/>
+      <Center>
+        <Form/>
+      </Center>
```
Rados13, line 11: Is it possible to center it vertically also?

Rados13 (approved):
I found only two small issues, but the whole PR looks good. Also, you can change PR status from draft to ready for review 😄

PR #4, opened 2022-07-19
LIVE-29/Feature/webinar gen json

pkrucz00:
Dodano operacje CRUD dla webinaru, czyli:
 - GET webinars
 - GET webinars/:id
 - PUT webinars/:id
 - POST webinars
 - DELETE webinars/:id

Uwagi::
- obecnie wyłączony jest CQRS (można włączyć w pliku `lib/router.ex:9`
- linki do webinarów to dla użytkownika `webinars/event/:SHA`, dla moderatora `webinars/events/:SHA/moderator`


lib/membrane_live/webinars/webinar.ex
```diff
@@ -0,0 +1,43 @@
...
+  import Ecto.Changeset
+
+  @type t :: %__MODULE__{
+          description: String.t(),
+          moderator_link: String.t(),
+          presenters: list,
+          start_date: NaiveDateTime.t(),
+          title: String.t(),
+          viewer_link: String.t()
+        }
```
Rados13, line 15: We could also think about generating viewer_link as `/event/<primary_key>`, this way we will not have to store it in DB as a separate field. But this way it is also good.

Example tutorial for setting it up https://elixircasts.io/uuid-primary-key-with-ecto.

Rados13 (changes requested):

lib/membrane_live/webinars.ex
```diff
@@ -0,0 +1,131 @@
...
+  @moduledoc """
+  The Webinars context.
+  """
+
+  import Ecto.Query, warn: false
+  alias MembraneLive.Repo
+
+  alias MembraneLive.Webinars.Webinar
+
+  @spec list_webinars :: any
```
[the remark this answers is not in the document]
pkrucz00: OK, changed most of `any` to more accurate typings

pkrucz00:
In my opinion - the `viewer` and `moderator_link` should be kept in the database (as was the original idea). They will be generated with uuid (not sha) and no `moderator` suffix will be added.

This way it will be easier for the backend to retrieve information and extend the database if needed and it would be harder for viewer to get the link of the moderator

lib/membrane_live/webinars.ex
```diff
@@ -0,0 +1,131 @@
...
+
+  Raises `Ecto.NoResultsError` if the Webinar does not exist.
+
+  ## Examples
+
+      iex> get_webinar!(123)
+      %Webinar{}
+
+      iex> get_webinar!(456)
+      ** (Ecto.NoResultsError)
```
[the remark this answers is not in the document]
pkrucz00: They'll be removed

lib/membrane_live/webinars.ex
```diff
@@ -0,0 +1,131 @@
...
+  """
+  def change_webinar(%Webinar{} = webinar, attrs \\ %{}) do
+    Webinar.changeset(webinar, attrs)
+  end
+
+  defp add_links(attrs) do
+    hash_code =
+      :crypto.hash(:sha256, "#{attrs["title"]}#{attrs["description"]}#{:rand.uniform(1_000_000)}")
+      |> Base.encode16()
+      |> String.downcase()
```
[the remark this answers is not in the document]
pkrucz00: UUID have been used.
I've addressed this issue in [this comment](https://github.com/membraneframework-labs/membrane_live/pull/4#issuecomment-1189887305)

lib/membrane_live/webinars/webinar.ex
pkrucz00: I've addressed this issue in [this comment](https://github.com/membraneframework-labs/membrane_live/pull/4#issuecomment-1189887305)

lib/membrane_live/webinars.ex
```diff
@@ -67,37 +48,13 @@ defmodule MembraneLive.Webinars do
...
-  @doc """
-  Updates a webinar.
-
-  ## Examples
-
-      iex> update_webinar(webinar, %{field: new_value})
-      {:ok, %Webinar{}}
-
-      iex> update_webinar(webinar, %{field: bad_value})
-      {:error, %Ecto.Changeset{}}
```
Rados13: Based on the example below, we can see that output of function will be `{:ok, %Webinar{}} | {:error, %Ecto.Changeset{}}`

lib/membrane_live/webinars/webinar.ex
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  @spec changeset(
+          {map, map}
+          | %{
+              :__struct__ => atom | %{:__changeset__ => map, optional(any) => any},
+              optional(atom) => any
+            },
+          :invalid | %{optional(:__struct__) => none, optional(atom | binary) => any}
+        ) :: Ecto.Changeset.t()
```
Rados13, line 35: As we removed dialyzer, you can go through files and remove these which aren't useful in your opinion.

Rados13 (approved):
These are only two small issues about typespecs

lib/membrane_live/webinars/webinar.ex
pkrucz00: OK, will do 🏃‍♂️

lib/membrane_live/webinars.ex
pkrucz00: Right 😅

lib/membrane_live/webinars/webinar.ex
pkrucz00: The problem with this is that `mix credo` pipeline won't pass because of public functions without `@spec`. Should I comment those functions with `#credo:ignore-next-line`? #️⃣

PR #10, opened 2022-07-21
Live 30/changing get for event page

pkrucz00:
Changes:
  - uuid as primary key in `webinars`
  - `GET webinars/:uuid ` response like the documentation
  - changes in credo - `@specs` are no longer checked
  - `alias` in `mix.ex`

### DIFF
--- a/test/membrane_live_web/controllers/webinar_controller_test.exs
+++ b/test/membrane_live_web/controllers/webinar_controller_test.exs
@@ -25,6 +25,9 @@ defmodule MembraneLiveWeb.WebinarControllerTest do
     "title" => nil
   }
 
+  @link_prefix "webinars/events/"
+  @moderator_link_suffix "/moderator"
+
   setup %{conn: conn} do
     {:ok, conn: put_req_header(conn, "accept", "application/json")}
   end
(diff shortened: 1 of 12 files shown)


test/membrane_live_web/controllers/webinar_controller_test.exs
```diff
@@ -44,21 +40,19 @@ defmodule MembraneLiveWeb.WebinarControllerTest do
...
       conn = post(conn, Routes.webinar_path(conn, :create), webinar: @create_attrs)
 
-      assert %{"viewer_link" => viewer_link, "moderator_link" => moderator_link} =
+      assert %{"viewer_link" => viewer_link, "moderator_link" => _moderator_link} =
                json_response(conn, 201)["webinar_links"]
 
-      webinar = Webinars.get_webinar_by_link(viewer_link)
-      id = webinar.id
+      uuid = get_uuid_from_link(viewer_link)
+      webinar = Webinars.get_webinar(uuid)
```
Rados13, line 47: We could add an assertion about whether `moderator_link` has a correct suffix and also whether both links have correct prefixes.

Rados13 (commented):
I know that these links are temporary, but still I would add tests for it to be sure that it works properly.

test/membrane_live_web/controllers/webinar_controller_test.exs
pkrucz00: Done 🏋️‍♂️

Rados13 (approved):

PR #11, opened 2022-07-21
Live 32 add form in event view

Karolk99:


lib/membrane_live_web/channels/event_socket.ex
```diff
@@ -0,0 +1,50 @@
...
+
+  # A Socket handler
+  #
+  # It's possible to control the websocket connection and
+  # assign values that can be accessed by your channel topics.
+
+  ## Channels
+  # Uncomment the following line to define a "room:*" topic
+  # pointing to the `MembraneLiveWeb.RoomChannel`:
+  #
```
Rados13, line 12: Do we need these comments?

assets/js/pages/Form.tsx
```diff
@@ -69,7 +69,7 @@ function Form() {
         return Promise.reject(response.status);
       })
       .then((data) => {
-        setLinks(JSON.parse(data).webinar_links);
+        setLinks(data.webinar_links);
```
Rados13, line 72: Could you make these links clickable and after clicking them redirect you to the proper page?

lib/membrane_live_web/channels/event_channel.ex
```diff
@@ -0,0 +1,27 @@
+defmodule MembraneLiveWeb.EventChannel do
+  use Phoenix.Channel
+  alias MembraneLiveWeb.Presence
+
+  def join("event:" <> _id, %{"name" => name}, socket) do
```
Rados13, line 5: Shouldn't we check in DB if an event with this id exists?

Rados13 (changes requested):
There is some bug regarding routing. When I paste a link, which I received after creating an event Phoenix returns an error that this route doesn't exist.

Rados13 (approved):
Nice job.

PR #12, opened 2022-07-22
LIVE-33/34: Retrieve list of participants and display it

LVala:
Added React component ParticipantsList which is a dynamic list of participants (currently with mocked photo and options button, nothing styled). Data about participants is retrieved from Presence via client side phoenix library.

### DIFF
--- a/assets/js/components/ParticipantsList.tsx
+++ b/assets/js/components/ParticipantsList.tsx
@@ -2,11 +2,8 @@ import React from "react";
 import {
   SimpleGrid,
   Box,
-  Table,
   Flex,
   Heading,
-  ButtonGroup,
-  Button,
   Spacer,
   Avatar,
   Menu,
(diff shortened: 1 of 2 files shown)


LVala:
Maybe it would be beneficial to write some tests for the channel/presence.

Rados13:
> Maybe it would be beneficial to write some tests for the channel/presence.

Yup I agree with that.

assets/js/components/ParticipantsList.tsx
```diff
@@ -0,0 +1,60 @@
...
+  IconButton,
+  MenuButton,
+  MenuList,
+  MenuItem,
+} from "@chakra-ui/react";
+
+const ParticipantMenu = (props) => {
+  return (
+    <Menu>
+      <MenuButton as={IconButton} aroa-label="Options" icon={"Menu Icon"} variant="outline" />
```
Rados13, line 22: I would change the color of this button to make it easier to find.

assets/js/components/ParticipantsList.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+const Participant = ({ name }) => {
+  return (
+    <Flex width="15vw" alignItems="center" gap="1">
+      <Avatar name="Dan Abrahmov" src="https://bit.ly/dan-abramov" />
+      <Box p="2">
+        <Heading size="md">{name}</Heading>
+      </Box>
+      <Spacer />
+      <ParticipantMenu />
```
Rados13, line 42: I would also think about that moderator and viewer will have different options available (I am not sure if viewer will have any options). So I would add some logic about changing options based on type of the user.

Rados13 (changes requested):
You could also add mentioned tests for Phoenix.Presence.

assets/js/components/ParticipantsList.tsx
LVala: I changed the color but everything style-wise is temporary anyway

assets/js/components/ParticipantsList.tsx
LVala: Added prop to the component telling if user is a moderator.

PR #14, opened 2022-07-26
Live 38/adding audio and video panel

pkrucz00:
Adding panel for choosing audio and video source


Rados13 (approved):
You can change the box's border which we mentioned in the previous talk.

pkrucz00:
I'll change it on the following branch LIVE-39

PR #15, opened 2022-07-26
LIVE-33: Add basic channel and presence tests

LVala:
Add basic channel and presence tests. At this point there's not much to test because the custom channel functionality have not been implemented yet.

### DIFF
--- a/test/membrane_live_web/channels/event_channel_test.exs
+++ b/test/membrane_live_web/channels/event_channel_test.exs
@@ -21,19 +21,20 @@ defmodule MembraneLiveWeb.EventChannelTest do
 
     {:ok, _reply, socket1} =
       subscribe_and_join(socket1, MembraneLiveWeb.EventChannel, "event:#{uuid}", %{
-        name: "Andrzej"
+        name: "John"
       })
 
     {:ok, _reply, socket2} =
-      subscribe_and_join(socket2, MembraneLiveWeb.EventChannel, "event:#{uuid}", %{name: "Janusz"})
+      subscribe_and_join(socket2, MembraneLiveWeb.EventChannel, "event:#{uuid}", %{name: "Adam"})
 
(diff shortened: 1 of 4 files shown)


test/membrane_live_web/channels/event_channel_test.exs
```diff
@@ -0,0 +1,43 @@
...
+
+    {:ok, _reply, socket1} =
+      subscribe_and_join(socket1, MembraneLiveWeb.EventChannel, "event:#{uuid}", %{
+        name: "Andrzej"
+      })
+
+    {:ok, _reply, socket2} =
+      subscribe_and_join(socket2, MembraneLiveWeb.EventChannel, "event:#{uuid}", %{name: "Janusz"})
+
+    assert %{"Andrzej" => %{}, "Janusz" => %{}} = Presence.list("event:#{uuid}")
```
Rados13, line 30: You didn't check here equality but assign value so it does nothing.

test/membrane_live_web/channels/event_channel_test.exs
```diff
@@ -0,0 +1,43 @@
...
+
+    Process.unlink(socket1.channel_pid)
+    close(socket1)
+
+    pres = Presence.list("event:#{uuid}")
+    assert is_map_key(pres, "Janusz") && !is_map_key(pres, "Andrzej")
```
Rados13, line 36: It is not C, you can use `and` and `not` 😃

Rados13 (changes requested):
I would also add a test of whether `Presence.update` works correctly.

test/membrane_live_web/channels/event_channel_test.exs
Rados13: Ok, I thought it works a different way.

Rados13 (approved):

PR #16, opened 2022-07-26
Live 35 add panel for choosing presenter

Karolk99:


assets/js/components/ParticipantsList.tsx
```diff
@@ -24,31 +34,56 @@ const ModeratorMenu = ({ isPresenter }) => {
...
-        <MenuItem>{isPresenter ? "Set as a normal participant" : "Set as a presenter"}</MenuItem>
-        <MenuItem>Mute</MenuItem>
-        <MenuItem>Kick</MenuItem>
+        <MenuItem
+          onClick={handleClick}
+          value={isPresenter ? "Set as a normal participant" : "Set as a presenter"}
+        >
+          {isPresenter ? "Set as a normal participant" : "Set as a presenter"}
+        </MenuItem>
+        <MenuItem onClick={handleClick} value="Mute">
```
Rados13, line 43: Why do you use the same `onclick` function for all menuItems?

Karolk99: Good point.

assets/js/components/PresenterPopup.tsx
```diff
@@ -7,17 +7,30 @@ import {
...
 
+  const onReject = () => {
+    eventChannel.push("presenter_answer", { name: name, moderator: moderator, answer: "reject" });
+    setPopupState({ isOpen: false, moderator: "" });
+  };
+
+  const onAccept = () => {
+    eventChannel.push("presenter_answer", { name: name, moderator: moderator, answer: "accept" });
+    setPopupState({ isOpen: false, moderator: "" });
+  };
```
Rados13, line 30: Could you move this to a separate function, which will take only the parameter `answer`?

assets/js/pages/Event.tsx
```diff
@@ -100,10 +103,28 @@ const Event = () => {
           </Button>
         </Box>
         <Spacer />
-        <ParticipantsList participants={participants} isModerator={eventInfo.is_moderator} />
+        <ParticipantsList
+          yourName={eventInfo.username}
+          eventChannel={eventChannel}
+          participants={participants}
+          isModerator={eventInfo.is_moderator}
```
Rados13, line 110: I think it would better look if we would stay in camelCase, probably you should map values after receiving value.

assets/js/pages/Form.tsx
```diff
@@ -12,74 +12,44 @@ import {
...
+export type EventForm = {
   title: string;
   description: string;
   start_date: string;
   presenters: string[];
 };
 
-type Links = {
+export type Links = {
   viewer_link: string;
```
mysza41, line 25: CamelCase to keep the consistency ;)

assets/js/pages/Form.tsx
```diff
@@ -12,74 +12,44 @@ import {
...
   moderator_link: string;
 };
 
 function Form() {
```
mysza41, line 29: Based on how many input fields and its states we have to handle here I would recommend to start using a true form :D And the easiest way to i implement and handle forms in react is to use dedicated library (e.g Formik or React-hook-form are great tools)
What I found  - chackra UI integrates with Formik in very nice way. Here example of use https://chakra-ui.com/getting-started/with-formik so to be considered ;)

mysza41 (commented):
So far so good.

PR #19, opened 2022-08-01
Live 42 spawn rtc engine

Karolk99:
When first user joins event phoenix spawns RTC Engine and HLS Endpoint also the user is able to connect to WebRTC Endpoint. Logic connected to presenters behaviour is not implemented in this pull request.


mix.exs
```diff
@@ -62,7 +61,32 @@ defmodule Membrane.Live.Mixfile do
...
+      {:opentelemetry, "~> 1.0"},
+      {:opentelemetry_api, "~> 1.0"},
+      {:opentelemetry_exporter, "~> 1.0"},
+      {:opentelemetry_zipkin, "~> 1.0"},
+
+      # Benchmarks
+      {:beamchmark, "~> 0.1.0", only: :benchmark},
+      {:stampede, github: "geometerio/stampede-elixir", only: :benchmark},
+      {:httpoison, "~> 1.8", only: :benchmark},
+      {:poison, "~> 5.0.0", only: :benchmark},
```
Rados13, line 80: You can remove these dependencies as you don't use any of them.

lib/membrane_live_web/channels/event_channel.ex
```diff
@@ -39,12 +42,60 @@ defmodule MembraneLiveWeb.EventChannel do
...
+        do_join(socket, event_pid, event_id)
+
+      {:error, reason} ->
+        Logger.error("""
+        Failed to start room.
+        Room: #{inspect(event_id)}
+        Reason: #{inspect(reason)}
+        """)
+
+        {:error, %{reason: "failed to start room"}}
```
Rados13, line 67: As we don't define the name Room in this context, I suggest changing Room to Event or Event stream.
```suggestion
        {:error, %{reason: "failed to start event"}}
```

lib/membrane_live/event.ex
```diff
@@ -0,0 +1,239 @@
...
+    Membrane.Logger.error("Endpoint #{inspect(endpoint_id)} has crashed!")
+    peer_channel = state.peer_channels[endpoint_id]
+
+    error_message = "WebRTC endpoint has crashed, please refresh the page to reconnect"
+    data = MediaEvent.create_error_event(error_message)
+    send(peer_channel, {:media_event, data})
+
+    {:noreply, state}
+  end
+
```
Rados13, line 179: As we could have different types you could change the error message and add some TODO as we probably will have handle different way for crash WebRTC than crash of HLS Endpoint.

lib/membrane_live/event.ex
```diff
@@ -0,0 +1,239 @@
...
+  @impl true
+  def handle_info({:playlist_playable, :audio, _playlist_idl}, state) do
+    {:noreply, state}
+  end
+
+  @impl true
+  def handle_info({:playlist_playable, :video, playlist_idl}, state) do
+    {:noreply, state}
+  end
```
Rados13: I would add some comment that this will be used for detecting when the HLS stream started.

Rados13 (approved):
The whole PR looks good, some small changes are required but these are mostly some comments or changing logging information.

PR #22, opened 2022-08-05
LIVE-41: Add WebRTC callbacks and self preview

LVala:
Add callbacks for the webrtc object from Mebmbrane client library, fix issue with joining multiple times from the same client and add self-preview (image from users camera).


assets/js/components/PresenterStreamArea.tsx
```diff
@@ -9,35 +9,36 @@ type PresenterStreamAreaProps = {
...
   }, [presenters]);
 
-  return (
-    <div>
+  return presenters.includes(username) ? (
+    <>
       {presenters.map((presenter) => {
         return (
           <RtcPlayer
+            username={username}
```
Rados13, line 31: IMO it's a very misleading name, as my first thought was that this is the username of the presenter. What do you think about changing it to `clientName` or even passing to the component flag whether the `isMe` or `isMyself` flag?

assets/js/pages/Event.tsx
```diff
@@ -95,9 +95,10 @@ const Event = () => {
   }, [name, eventChannel]);
 
   useEffect(() => {
-    const alreadyJoined = privateChannel?.state === "joined";
-    if (name && !alreadyJoined) {
-      const channel = socket.channel(`private:${eventInfo.link}:${name}`, {});
+    const privateAlreadyJoined = privateChannel?.state === "joined";
+    const eventAlreadyJoined = eventChannel?.state === "joined";
+    if (name && !privateAlreadyJoined && eventAlreadyJoined) {
+      const channel = socket.channel("private:" + eventInfo.link + ":" + name, {});
```
Rados13, line 101: Why do you change template literals to the concatenation of strings?

assets/js/utils/rtcUtils.ts
```diff
@@ -37,31 +37,51 @@ export const connectWebrtc = async (
     console.error("Couldn't get camera permission:", error);
   }
 
+  const updateStream = (toAdd: boolean, stream: MediaStream | null, name: string) => {
+    // adds new stream when toAdd is true, else removes the stream
+    if (toAdd && stream != null) presenterStreams[name] = stream;
+    let new_elem = {};
+    new_elem[name] = true;
+    setStreamsAvailable({ ...streamsAvailable, ...new_elem });
+  };
```
Rados13, line 46: Why are we doing this in one function? What is the problem with creating `addStream` and `removeStream` functions?

assets/js/utils/rtcUtils.ts
```diff
@@ -37,31 +37,51 @@ export const connectWebrtc = async (
     console.error("Couldn't get camera permission:", error);
   }
 
+  const updateStream = (toAdd: boolean, stream: MediaStream | null, name: string) => {
+    // adds new stream when toAdd is true, else removes the stream
+    if (toAdd && stream != null) presenterStreams[name] = stream;
+    let new_elem = {};
+    new_elem[name] = true;
+    setStreamsAvailable({ ...streamsAvailable, ...new_elem });
```
Rados13, line 45: This `new_elem` variable is useless.
```suggestion
    let new_elem = {};
    new_elem[name] = true;
    setStreamsAvailable({ ...streamsAvailable, ...{[name]: true}});
```

assets/js/utils/rtcUtils.ts
```diff
@@ -70,5 +90,26 @@ export const connectWebrtc = async (
...
+  webrtc: any,
+  name: string,
+  streamsAvailable: { [key: string]: boolean },
+  setStreamsAvailable: React.Dispatch<React.SetStateAction<{ [key: string]: boolean }>>,
+  webrtcChannel: any
+) => {
+  webrtcChannel.off("mediaEvent");
+  let new_elem = {};
+  new_elem[name] = false;
+  setStreamsAvailable({ ...streamsAvailable, ...new_elem });
```
Rados13, line 110: Same as above.

assets/js/utils/rtcUtils.ts
```diff
@@ -70,5 +90,26 @@ export const connectWebrtc = async (
...
-  webrtcChannel.on("mediaEvent", (event) => webrtc.receiveMediaEvent(event.data));
+  webrtcChannel.on("mediaEvent", (event) => {
+    webrtc.receiveMediaEvent(event.data);
+  });
+
+  return webrtc;
+};
+
+export const leaveWebrtc = async (
+  webrtc: any,
```
Rados13, line 101: Shouldn't it be type `MembraneWebRTC`?

Rados13 (changes requested):

Rados13:
Also, buttons for controlling media during WebRTC calls are still disabled.

assets/js/components/PresenterStreamArea.tsx
```diff
@@ -1,34 +1,39 @@
...
+}: PresenterStreamAreaProps) => {
   const [streamsAvailable, setStreamsAvailable] = useState<{ [key: string]: boolean }>({});
 
   useEffect(() => {
-    if (webrtc == null && presenters.includes(username)) {
-      webrtc = connectWebrtc(eventChannel, username, streamsAvailable, setStreamsAvailable);
-    } else if (webrtc != null && !presenters.includes(username)) {
-      leaveWebrtc(webrtc, username, streamsAvailable, setStreamsAvailable, eventChannel);
+    if (webrtc == null && presenters.includes(clientName)) {
+      webrtc = connectWebrtc(eventChannel, clientName, streamsAvailable, setStreamsAvailable);
```
Rados13, line 24: Why does it return a promise? Could we make it return `MembraneWebrtc` or use `then` and in then assign value to webrtc.

assets/js/components/RtcPlayer.tsx
```diff
@@ -1,13 +1,13 @@
...
 type RtcPlayerProps = {
-  username: string;
+  clientName: string;
   name: string;
   presenterStreams: { [key: string]: MediaStream };
   streamsAvailable: { [key: string]: boolean };
 };
 
-const RtcPlayer = ({ username, name, presenterStreams, streamsAvailable }: RtcPlayerProps) => {
+const RtcPlayer = ({ clientName, name, presenterStreams, streamsAvailable }: RtcPlayerProps) => {
```
Rados13, line 10: As I look at that a second time I see that we use this variable only for checking whether a name is an equal client, so maybe it will be better to pass a boolean, instead of a string.

Rados13 (approved):

PR #23, opened 2022-08-08
Live 36 temporary handle a change of presenter

Karolk99:
Includes LIVE-40 (Display username of presenter).


lib/membrane_live/event.ex
```diff
@@ -165,13 +164,33 @@ defmodule MembraneLive.Event do
     Engine.accept_peer(rtc_engine, peer.id)
     :ok = Engine.add_endpoint(rtc_engine, endpoint, peer_id: peer.id, node: peer_node)
 
+    state = update_in(state, [:playlist_idls], fn items -> [%{peer_id: peer.id} | items] end)
```
Rados13, line 167: As you are using only one key, you can use simply `Map.update/3`.

lib/membrane_live/event.ex
```diff
@@ -165,13 +164,33 @@ defmodule MembraneLive.Event do
...
+      state.playlist_idls == [] ->
+        MembraneLiveWeb.Endpoint.broadcast!("event:" <> state.event_id, "playlist_playable", %{
+          playlist_idl: ""
+        })
+
+      hd(state.playlist_idls).playlist_idl != prev_playlist_idl ->
+        MembraneLiveWeb.Endpoint.broadcast!("event:" <> state.event_id, "playlist_playable", %{
+          playlist_idl: hd(state.playlist_idls).playlist_idl
+        })
+    end
```
Rados13, line 192: This is complicated and vulnerable to race conditions, so I would advise extending the HLSEndpoint message `:playlist_playable`, so it would contain also the origin of the track.

Rados13 (changes requested):

lib/membrane_live/event.ex
```diff
@@ -87,13 +87,12 @@ defmodule MembraneLive.Event do
 
     {:ok,
      %{
-       playlist_idl: "",
+       playlist_idls: [],
```
Rados13, line 90: Why this is a list?

lib/membrane_live/event.ex
```diff
@@ -216,18 +226,25 @@ defmodule MembraneLive.Event do
...
-    state = put_in(state, [:playlist_idl], playlist_idl)
+  def handle_info({:playlist_playable, :video, playlist_idl, peer_id}, state) do
+    state =
+      Map.put(
+        state,
+        :playlist_idls,
+        Enum.map(state.playlist_idls, &add_playlist_idl_to_peer(&1, playlist_idl, peer_id))
+      )
+
+    name = Enum.find(state.playlist_idls, fn map -> map[:playlist_idl] == playlist_idl end).name
```
Rados13, line 243: Why don't we simply store under in `state.playlist_idls` a map `%{ peer_id => {peer_name, playlist_idl}}`? For deciding which stream should be played now, we could simply use a queue or stack of `peer_ids`.

lib/membrane_live/event.ex
```diff
@@ -241,8 +258,17 @@ defmodule MembraneLive.Event do
...
+        %{playlist_idl: playlist_idl, name: name} ->
+          {:reply, %{playlist_idl: playlist_idl, name: name}, state}
+
+        _no_playlist_idl ->
+          {:reply, %{playlist_idl: "", name: ""}, state}
+      end
+    else
+      {:reply, %{playlist_idl: "", name: ""}, state}
+    end
   end
```
Rados13, line 272: To make it simplier
```suggestion
      case state.playlist_idls do
        [%{playlist_idl: playlist_idl, name: name} | _tail] ->
          {:reply, %{playlist_idl: playlist_idl, name: name}, state}
         _other -> {:reply, %{playlist_idl: "", name: ""}, state}
      end
  end
```

lib/membrane_live/event.ex
```diff
@@ -253,4 +279,36 @@ defmodule MembraneLive.Event do
     ]
 
   defp event_span_id(id), do: "event:#{id}"
+
+  defp add_playlist_idl_to_peer(peer_hls, playlist_idl, peer_id) do
+    if peer_hls.peer_id == peer_id,
+      do: Map.put(peer_hls, :playlist_idl, playlist_idl),
+      else: peer_hls
+  end
```
Rados13, line 287: IMO it's pretty strange that function `add_playlist_idl_to_peer` doesn't always add playlist_idl to peer.

lib/membrane_live/event.ex
```diff
@@ -253,4 +279,36 @@ defmodule MembraneLive.Event do
...
+    cond do
+      state.playlist_idls == [] ->
+        MembraneLiveWeb.Endpoint.broadcast!("event:" <> state.event_id, "playlist_playable", %{
+          playlist_idl: "",
+          name: ""
+        })
+
+      hd(state.playlist_idls).playlist_idl != prev_playlist_idl ->
+        MembraneLiveWeb.Endpoint.broadcast!("event:" <> state.event_id, "playlist_playable", %{
+          playlist_idl: hd(state.playlist_idls).playlist_idl,
```
Rados13, line 304: You could move it to the helper variable or probably use `prev_playlist_idl`.

Rados13 (changes requested):

lib/membrane_live/event.ex
```diff
@@ -280,33 +276,28 @@ defmodule MembraneLive.Event do
...
-    if peer_hls.peer_id == peer_id,
-      do: Map.put(peer_hls, :playlist_idl, playlist_idl),
-      else: peer_hls
-  end
-
   defp handle_peer_left(state, peer_id) do
-    prev_playlist_idl = hd(state.playlist_idls).playlist_idl
+    [prev_peer | _rest] = state.peer_ids
+    state = update_in(state, [:playlist_idls], &Map.delete(&1, peer_id))
+    state = update_in(state, [:peer_ids], &Enum.reject(&1, fn id -> id == peer_id end))
```
Rados13, line 282: You can use `Map.update!/3` as we use only one key.

Rados13 (approved):

PR #25, opened 2022-08-09
LIVE-78: Connect RTC Player with media source controls

LVala:
Connect RTC Player Player with media source controls component (add changing and muting media device). Currently it is expected that the user will choose a device after joining.

### DIFF
--- a/assets/js/components/PresenterStreamArea.tsx
+++ b/assets/js/components/PresenterStreamArea.tsx
@@ -1,7 +1,9 @@
 import React, { useEffect, useState } from "react";
-import { connectWebrtc, leaveWebrtc } from "../utils/rtcUtils";
+import { connectWebrtc, leaveWebrtc, replaceTrack } from "../utils/rtcUtils";
 import { MembraneWebRTC } from "@membraneframework/membrane-webrtc-js";
-import RtcPlayer from "./RtcPlayer";
+import { RtcPlayer, RtcClientPlayer } from "./RtcPlayer";
+import type { SourceType } from "./ControlPanel";
+import ControlPanel from "./ControlPanel";
 
 type PresenterStreamAreaProps = {
   clientName: string;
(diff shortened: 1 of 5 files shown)


assets/js/components/PresenterStreamArea.tsx
```diff
@@ -12,12 +14,59 @@ type PresenterStreamAreaProps = {
 export const presenterStreams: { [key: string]: MediaStream } = {};
 let webrtc: MembraneWebRTC | null = null;
 
+export const changeIsTrackEnabled = (clientName: string, sourceType: SourceType, mode: boolean) => {
+  const track = presenterStreams[clientName].getTracks().find((track) => track.kind == sourceType);
```
Rados13, line 18: We repeat this fragment so we can move it to the helper function.

assets/js/components/PresenterStreamArea.tsx
```diff
@@ -12,12 +14,59 @@ type PresenterStreamAreaProps = {
...
+) => {
+  if (presenterStreams[clientName] == undefined) presenterStreams[clientName] = new MediaStream();
+  else {
+    const curTrack = presenterStreams[clientName]
+      .getTracks()
+      .find((track) => track.kind == sourceType);
+    if (curTrack != undefined) {
+      curTrack.stop();
+      presenterStreams[clientName].removeTrack(curTrack);
+    }
```
Rados13, line 37: According to https://developer.mozilla.org/en-US/docs/Web/API/MediaStream nothing happens if the passed track doesn't exist in the media stream.
```suggestion
      curTrack?.stop();
      presenterStreams[clientName].removeTrack(curTrack);
    }
```

Rados13, line 37: Why do we remove tracks in `addMediaStreamTrack`?

assets/js/components/PresenterStreamArea.tsx
```diff
@@ -32,19 +81,26 @@ const PresenterStreamArea = ({
...
+          return (
+            <RtcPlayer
+              isMyself={clientName == presenter}
+              name={presenter}
+              streamsAvailable={streamsAvailable}
+              key={presenter}
+            />
+          );
       })}
+      <ControlPanel clientName={clientName} setLastSourceTypeChange={setLastSourceTypeChange} />
```
Rados13, line 103: Wouldn't it be more readable to first filter presenters and then map? We could also think about storing them separately in the state?

assets/js/components/RtcPlayer.tsx
```diff
@@ -26,10 +27,32 @@ const RtcPlayer = ({ isMyself, name, presenterStreams, streamsAvailable }: RtcPl
...
+type RtcClientPlayerProps = {
+  name: string;
+  clientStreamAvailable: SourceType;
+};
+
+export const RtcClientPlayer = ({ name, clientStreamAvailable }: RtcClientPlayerProps) => {
+  const videoRef = useRef<HTMLVideoElement>(null);
+
+  useEffect(() => {
+    if (name in presenterStreams && clientStreamAvailable && videoRef.current != null) {
```
Rados13, line 45: It's a pretty complex condition, move it to a helper function.

assets/js/utils/rtcUtils.ts
```diff
@@ -62,44 +38,61 @@ export const connectWebrtc = async (
...
 };
 
+export const replaceTrack = (
+  webrtc: MembraneWebRTC | null,
+  clientName: string,
+  mediaStream: MediaStream,
+  sourceType: SourceType
+) => {
+  const track = mediaStream.getTracks().find((track) => track.kind == sourceType);
+  if (webrtc != null && track != undefined) {
```
Rados13, line 89: Negate this condition and make a return in its body. It will remove one level of nesting.

Rados13 (changes requested):

assets/js/components/PresenterStreamArea.tsx
LVala: But that's not the case here, argument passed to removeTrack must not be undefined (or am I missing something?)

LVala: I don't want client MediaStream to have more than 1 track of each kind (audio/video), that wasn't tested tho (because of lack of more than 1 actual real devices).

assets/js/components/PresenterStreamArea.tsx
LVala: Filtering - fixed. "presenters" is already a state passed down from Event component as a prop.

assets/js/utils/rtcUtils.ts
LVala: Fixed

assets/js/components/PresenterStreamArea.tsx
LVala: Fixed

assets/js/components/RtcPlayer.tsx
LVala: I honestly thing it would add even more complexity

assets/js/components/PresenterStreamArea.tsx
LVala: Done

Rados13 (approved):

PR #26, opened 2022-08-10
Live 72 live 71 interceptor and redirect

pkrucz00:
Client side authentication logic. Includes features like:
 - redirecting using `GuardedRoute`
 - sending jwt in header via the interceptor
 - clearing localStorage "jwt" value after receiving 401


assets/js/pages/Auth.tsx
```diff
@@ -1,25 +1,26 @@
...
-    const request: RequestInit = {
-      method: "POST",
-      headers: { "Content-Type": "application/json" },
-      body: JSON.stringify(google_response),
-    };
-
-    fetch(window.location.origin + "/auth", request)
-      .then((resp) => resp.json())
-      .then((data) => localStorage.setItem("jwt", data["token"]));
+  const getToken = async (google_response) => {
```
Rados13, line 10: Variable name should be in camelCase.

assets/js/pages/Auth.tsx
```diff
@@ -1,25 +1,26 @@
...
+    try {
+      const response = await axios.post("auth", google_response);
+      localStorage.setItem("jwt", response.data.token);
+    } catch (error) {
+      alert("Couldn't get the token. Please try again in a moment");
+    }
```
Rados13, line 16: Shouldn't we add some logs about what error it was?

assets/js/pages/Auth.tsx
Rados13, line 10: Shouldn't function getToken, return token?

assets/js/services/GuardedRoute.tsx
```diff
@@ -0,0 +1,9 @@
+import React from "react";
+import { Navigate, Outlet } from "react-router-dom";
+
+const GuardedRoute = () => {
+  const auth = localStorage.getItem("jwt") != null;
```
Rados13, line 5: - I would change the variable name to `isAuthenticated`
- I would move it to function like `isUserAuthenticated` or something like this. It could be in file `services/index.js.

lib/membrane_live_web/router.ex
```diff
@@ -4,6 +4,7 @@ defmodule MembraneLiveWeb.Router do
   pipeline :browser do
     plug(:accepts, ["json"])
     plug(:put_root_layout, {MembraneLiveWeb.LayoutView, :root})
+    # plug(MembraneLiveWeb.Plugs.Auth)
```
Rados13, line 7: Hmmm?

Rados13 (changes requested):

assets/js/pages/Auth.tsx
pkrucz00: 🐍➡️🐪

assets/js/pages/Auth.tsx
pkrucz00: Will simple `console.log(error);` do?

assets/js/pages/Auth.tsx
pkrucz00: Right. I've changed it to `fetchToken`

assets/js/services/GuardedRoute.tsx
pkrucz00: Done ✅

lib/membrane_live_web/router.ex
pkrucz00: Yep, it should

pkrucz00: Moved to mixed 🗑

assets/js/pages/Auth.tsx
pkrucz00: Done ✅

Rados13 (approved):

Rados13 (changes requested):
It is old code fix it.

pkrucz00:
Branch is for deletion. Cause: The git history is too convoluted to comprehend. The current branch for this is [this one](https://github.com/membraneframework-labs/membrane_live/tree/LIVE-71-LIVE-72-interceptor)

PR #27, opened 2022-08-11
LIVE-66: Add bearer/accessing jwt plug

LVala:


lib/membrane_live_web/controllers/fallback_controller.ex
```diff
@@ -15,11 +15,19 @@ defmodule MembraneLiveWeb.FallbackController do
     |> render("error.json", changeset: changeset)
   end
 
-  # # This clause is an example of how to handle resources that cannot be found.
+  # This clause is an example of how to handle generic
```
pkrucz00, line 18: What does it mean?

lib/membrane_live_web/plugs/bearer_plug.ex
```diff
@@ -0,0 +1,26 @@
...
+  import Plug.Conn
+  alias MembraneLive.Tokens
+  alias MembraneLiveWeb.FallbackController
+
+  def init(default), do: default
+
+  def call(conn, _default) do
+    case Enum.find(conn.req_headers, fn elem -> match?({"bearer", _}, elem) end) do
+      {"bearer", jwt} ->
+        %{"user_id" => user_id} = Tokens.custom_decode(jwt)
```
pkrucz00, line 15: - the header key we are looking for is `authorization`
 - the header vallue is always "Bearer {jwt}", so we need to use a regex

pkrucz00 (commented):
The bearer auth has a little different api, so it would be neat if you could change it

pkrucz00 (approved):

PR #36, opened 2022-08-17
Add header component

Karolk99:


assets/js/utils/headerUtils.ts
```diff
@@ -0,0 +1,47 @@
...
+    start_date: "",
+    presenters: [],
+  };
+};
+
+export const getEventInfo = (setEventInfo: React.Dispatch<React.SetStateAction<EventInfo>>) => {
+  axios
+    .get("/resources/webinars/" + getChannelId())
+    .then((response) => {
+      console.log(response.data);
```
Rados13, line 20: Do we need this?

assets/js/utils/headerUtils.ts
```diff
@@ -0,0 +1,47 @@
...
+      const start_date = response.data.webinar.start_date.replace("T", " ");
+      setEventInfo({ ...response.data.webinar, start_date: start_date });
+    })
+    .catch(() => {
+      alert("Couldn't get event information. Please reload this page.");
+    });
```
Rados13, line 26: Maybe we should log this error?

assets/js/pages/Event.tsx
```diff
@@ -72,16 +36,12 @@ const Event = () => {
...
-  }, []);
-
   useEffect(() => {
     const alreadyJoined = eventChannel?.state === "joined";
     if (name && !alreadyJoined) {
-      const channel = socket.channel(`event:${eventInfo.link}`, {
+      const channel = socket.channel(`event:${getChannelId()}`, {
         name: name,
-        isModerator: eventInfo.isModerator,
+        isModerator: true,
```
Rados13, line 44: Are you sure that this should be always true?

assets/js/pages/Event.tsx
```diff
@@ -108,11 +70,7 @@ const Event = () => {
           </div>
         </div>
         <div className="Participants">
-          <ParticipantsList
-            username={name}
-            isModerator={eventInfo.isModerator}
-            eventChannel={eventChannel}
-          />
+          <ParticipantsList username={name} isModerator={true} eventChannel={eventChannel} />
```
Rados13, line 73: Same as above.

assets/js/utils/channelUtils.ts
```diff
@@ -102,3 +103,11 @@ export const syncPresenters = (
...
   }
 };
+
+export const getChannelId = (): string => window.location.pathname.split("/")[2];
+
+const compareParticipants = (x: Participant, y: Participant): number => {
+  if (x.isModerator != y.isModerator) return x.isModerator ? -1 : 1;
+  if (x.isPresenter != y.isPresenter) return x.isPresenter ? -1 : 1;
+  return 0;
+};
```
Rados13, line 113: I think it is too complex I would create a function e.g: `participantToNumber` which takes the participant and return rank of this participant. Then this function could look like this:
```suggestion
const compareParticipants = (x: Participant, y: Participant): number => {
return participantToNumber(x) - participantToNumber(y)
};
```

assets/js/utils/headerUtils.ts
```diff
@@ -0,0 +1,47 @@
...
+  if (eventChannel) {
+    const presence = new Presence(eventChannel);
+
+    const updateParticipantsNumber = () => {
+      let counter = 0;
+      presence.list((name: string, metas: any) => {
+        counter += 1;
+      });
+      setParticipantsNumber(counter);
+    };
```
Rados13, line 39: I think this way it is simpler.
```suggestion
    const updateParticipantsNumber = () => {
      setParticipantsNumber(presence.list().length);
    };
```

Rados13 (changes requested):

assets/js/pages/Event.tsx
Karolk99: We don't have eventInfo state in Event component, isModerator will be added in the next pull request. Right now value is mocked.

assets/js/pages/Event.tsx
Karolk99: same answer

assets/js/utils/headerUtils.ts
```diff
@@ -0,0 +1,42 @@
...
+};
+
+export const getEventInfo = (setEventInfo: React.Dispatch<React.SetStateAction<EventInfo>>) => {
+  axios
+    .get("/resources/webinars/" + getChannelId())
+    .then((response) => {
+      const start_date = response.data.webinar.start_date.replace("T", " ");
+      setEventInfo({ ...response.data.webinar, start_date: start_date });
+    })
+    .catch(() => {
```
Rados13, line 23: Log error which we receive in catch.

Rados13 (approved):

PR #37, opened 2022-08-17
LIVE-66: Detect moderator based on authentication data

LVala:
Add modifications to "webinars" database table, functionality to save moderator name in DB and retrieve it when joining webinar channel.


LVala:
This functionality was reimplemented (based on what is here) in #41 .

PR #46, opened 2022-08-24
Add persistent presenter

Karolk99:
At this moment problem with multiple windows for one user i solved by not allowing user to connect to channel when he is already connected in other window.

### DIFF
--- a/assets/js/components/PresenterStreams.tsx
+++ b/assets/js/components/PresenterStreams.tsx
@@ -15,7 +15,8 @@ type PresenterStreamAreaProps = {
 };
 
 const playerCallbacks: { [key: string]: (sourceType: SourceType) => void } = {};
-let webrtc: MembraneWebRTC | null | boolean = null;
+let webrtc: MembraneWebRTC | null = null;
+let webrtcConnecting: boolean = false;
 
 const PresenterStreams = ({
   clientName,
@@ -27,14 +28,15 @@ const PresenterStreams = ({
   const [isControlPanelAvailable, setIsControlPanelAvailable] = useState(false);
 
   useEffect(() => {
-    if (webrtc == null && presenters.includes(clientName)) {
-      webrtc = true;
+    if (!webrtcConnecting && webrtc == null && presenters.includes(clientName)) {
+      webrtcConnecting = true;
       connectWebrtc(eventChannel, clientName, playerCallbacks).then((value) => {
         webrtc = value;
         setIsControlPanelAvailable(true);
+        webrtcConnecting = false;
       });
     } else if (webrtc != null && !presenters.includes(clientName)) {
-      leaveWebrtc(webrtc as MembraneWebRTC, clientName, eventChannel);
+      leaveWebrtc(webrtc, clientName, eventChannel);
       webrtc = null;
       setIsControlPanelAvailable(false);
     }
--- a/assets/js/components/PresenterStreams.tsx
+++ b/assets/js/components/PresenterStreams.tsx
@@ -15,7 +15,7 @@ type PresenterStreamAreaProps = {
 };
 
 const playerCallbacks: { [key: string]: (sourceType: SourceType) => void } = {};
-let webrtc: MembraneWebRTC | null = null;
+let webrtc: MembraneWebRTC | null | boolean = null;
 
 const PresenterStreams = ({
   clientName,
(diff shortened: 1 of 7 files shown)


lib/membrane_live_web/channels/event_channel.ex
```diff
@@ -17,26 +17,49 @@ defmodule MembraneLiveWeb.EventChannel do
...
-          Presence.track(socket, name, %{is_moderator: is_moderator, is_presenter: false})
-          {:ok, %{is_moderator: is_moderator}, socket}
-        else
-          _error ->
-            {:error,
-             %{reason: "Error occured while creating event stream or adding user to presence"}}
+  def join("event:" <> id, %{"token" => token, "reloaded" => reloaded}, socket) do
+    case Ecto.UUID.cast(id) do
+      {:ok, id} ->
+        case Repo.exists?(from(w in Webinar, where: w.uuid == ^id)) do
```
Rados13, line 23: I think that this two cases should be merged.

lib/membrane_live_web/channels/event_channel.ex
```diff
@@ -81,6 +104,30 @@ defmodule MembraneLiveWeb.EventChannel do
     {:ok, Phoenix.Socket.assign(socket, %{peer_id: peer_id})}
   end
 
+  defp check_if_presenter(name, reloaded, id) do
+    [{_key, presenters}] = :ets.lookup(:presenters, id)
+    in_ets = if Enum.find(presenters, &(&1 == name)) != nil, do: true, else: false
```
Rados13, line 109: I think this if is useless here.
```suggestions
in_ets = Enum.find(presenters, &(&1 == name)) != nil
```

lib/membrane_live_web/channels/event_channel.ex
```diff
@@ -81,6 +104,30 @@ defmodule MembraneLiveWeb.EventChannel do
...
+
+    case reloaded do
+      true ->
+        {:ok, in_ets}
+
+      false ->
+        if in_ets, do: remove_from_presenters(name, id)
+        {:ok, false}
+    end
```
Rados13, line 118: I think we can fully use case here, instead of switching if to case. 
```suggestions
    case {reloaded, in_ets} do
      {true, _in_ets} -> {:ok, in_ets}
      {false, true} ->
        remove_from_presenters(name, id)
        {:ok, false}
      {false, false} -> {:ok, false}
    end
```

Rados13 (changes requested):

assets/js/components/PresenterStreams.tsx
```diff
@@ -15,7 +15,7 @@ type PresenterStreamAreaProps = {
 };
 
 const playerCallbacks: { [key: string]: (sourceType: SourceType) => void } = {};
-let webrtc: MembraneWebRTC | null = null;
```
Rados13, line 18: This solution with boolean was mostly for debugging reasons. It is a hack and should be changed.

assets/js/utils/storageUtils.ts
```diff
@@ -47,3 +48,10 @@ export const storageSetEmail = (email: string): void => {
 export const storageSetPicture = (picture: string): void => {
   localStorage.setItem(PICTURE, picture);
 };
+
+export const storageSetReloaded = (): void => {
+  sessionStorage.setItem(RELOAD, "true");
+};
+export const storageGetReloaded = (): boolean => {
+  return sessionStorage.getItem(RELOAD) ? true : false;
```
Rados13, line 56:
```suggestion
  return sessionStorage.getItem(RELOAD) === "true";
```

assets/js/components/PresenterStreams.tsx
```diff
@@ -27,13 +28,15 @@ const PresenterStreams = ({
...
+    if (!webrtcConnecting && webrtc == null && presenters.includes(clientName)) {
+      webrtcConnecting = true;
       connectWebrtc(eventChannel, clientName, playerCallbacks).then((value) => {
         webrtc = value;
         setIsControlPanelAvailable(true);
+        webrtcConnecting = false;
       });
     } else if (webrtc != null && !presenters.includes(clientName)) {
-      leaveWebrtc(webrtc, clientName, eventChannel);
+      leaveWebrtc(webrtc as MembraneWebRTC, clientName, eventChannel);
```
Rados13, line 39: Probably not null casting here will be enough.

Rados13 (approved):

PR #47, opened 2022-08-25
Fix missing audio header

Karolk99:


Rados13 (commented):
Shouldn't we also update `rtc_engine` dependencies?

Rados13:
> Shouldn't we also update `rtc_engine` dependencies?

Bump.

Karolk99:
> > Shouldn't we also update `rtc_engine` dependencies?

> Bump.

Right now there is updated dependencies. Before it wasn't possible because there hasn't been  any update.

Rados13 (approved):

PR #51, opened 2022-08-30
Live 98 presence key email

Karolk99:


assets/js/components/PresenterStreams.tsx
```diff
@@ -18,25 +25,20 @@ const playerCallbacks: { [key: string]: (sourceType: SourceType) => void } = {};
...
-  setMode,
-}: PresenterStreamAreaProps) => {
-  const [presenters, setPresenters] = useState<string[]>([]);
+const PresenterStreams = ({ client, eventChannel, mode, setMode }: PresenterStreamAreaProps) => {
+  const [presenters, setPresenters] = useState<Presenter[]>([]);
   const [isControlPanelAvailable, setIsControlPanelAvailable] = useState(false);
 
   useEffect(() => {
-    if (!webrtcConnecting && webrtc == null && presenters.includes(clientName)) {
+    if (!webrtcConnecting && webrtc == null && presenters.some((e) => e.email === client.email)) {
```
Rados13, line 33: We use `presenters.some((e) => e.email == client.email` in both conditions. Move it to variable.

assets/js/components/PresenterStreams.tsx
```diff
@@ -46,26 +48,26 @@ const PresenterStreams = ({
...
-  return presenters.includes(clientName) ? (
+  return presenters.some((e) => e.name === client.name) ? (
     <div className={`PresenterStreams ${mode == "hls" ? "Hidden" : ""}`}>
       <div className={`StreamsGrid Grid${presenters.length}`}>
         {presenters.map((presenter) => {
           return (
             <RtcPlayer
-              isMyself={clientName == presenter}
-              name={presenter}
+              isMyself={client == presenter}
```
Rados13, line 57: Wouldn't it fail as `client` and `presenter` has different number of fields?

assets/js/components/PresenterStreams.tsx
```diff
@@ -46,26 +48,26 @@ const PresenterStreams = ({
...
+              presenter={presenter}
               playerCallbacks={playerCallbacks}
-              key={presenter}
+              key={presenter.name}
```
Rados13, line 60: Where do we use this `key`?

assets/js/components/RtcPlayer.tsx
```diff
@@ -1,29 +1,30 @@
...
+const RtcPlayer = ({ isMyself, presenter, playerCallbacks }: RtcPlayerProps) => {
   const videoRef = useRef<HTMLVideoElement>(null);
   const audioRef = useRef<HTMLAudioElement>(null);
 
   const style = getComputedStyle(document.body);
   const font = style.getPropertyValue("--font-light-color");
 
   const connectStreams = (sourceType: SourceType) => {
-    if (!(name in presenterStreams)) return;
+    if (!(presenter.email in presenterStreams)) return;
```
Rados13, line 21:
```suggestion
    if (!presenterStreams.includes(presenter.email)) return;
```

lib/membrane_live/event.ex
```diff
@@ -188,13 +188,17 @@ defmodule MembraneLive.Event do
...
   def handle_info(%Message.EndpointCrashed{endpoint_id: endpoint_id}, state) do
     Membrane.Logger.error("Endpoint #{inspect(endpoint_id)} has crashed!")
-    peer_channel = state.peer_channels[endpoint_id]
 
-    error_message = "Endpoint has crashed."
-    data = MediaEvent.create_error_event(error_message)
-    send(peer_channel, {:media_event, data})
+    case state.peer_channels[endpoint_id] do
+      nil ->
+        {:noreply, state}
```
Rados13, line 194: Add some log that this peer doesn't exist already.

lib/membrane_live_web/channels/event_channel.ex
```diff
@@ -163,17 +165,17 @@ defmodule MembraneLiveWeb.EventChannel do
...
     "event:" <> id = socket.topic
-    remove_from_presenters(presenter, id)
+    remove_from_presenters(email, id)
     {:noreply, socket}
   end
 
-  def handle_in("presenter_remove", %{"presenter_topic" => presenter_topic}, socket) do
-    props = Presence.get_by_key(socket, List.last(String.split(presenter_topic, ":")))
+  def handle_in("presenter_remove", %{"presenterTopic" => presenter_topic}, socket) do
+    presenter_email = List.last(String.split(presenter_topic, ":"))
```
Rados13, line 177: According to the style guide, which we follow in Membrane you should use pipe in situations like this.
https://github.com/christopheradams/elixir_style_guide#pipe-operator

lib/membrane_live_web/channels/event_channel.ex
```diff
@@ -189,32 +191,38 @@ defmodule MembraneLiveWeb.EventChannel do
...
         "presenter_answer",
-        %{"answer" => answer, "name" => name, "moderator" => moderator},
+        %{"answer" => answer, "email" => email, "moderatorTopic" => moderator_topic},
         socket
       ) do
     {:ok, socket} =
       if answer == "accept" do
         {:ok, _ref} =
-          Presence.update(socket, name, fn map -> Map.put(map, :is_presenter, true) end)
+          Presence.update(socket, email, fn map -> Map.put(map, :is_presenter, true) end)
```
Rados13, line 214: I think we should change map name or use lambda with `&` shorthand.

lib/membrane_live_web/channels/event_channel.ex
```diff
@@ -189,32 +191,38 @@ defmodule MembraneLiveWeb.EventChannel do
...
         "event:" <> id = socket.topic
-        add_to_presenters(name, id)
+        add_to_presenters(email, id)
         join_event_stream(socket)
       else
         {:ok, socket}
       end
 
-    MembraneLiveWeb.Endpoint.broadcast_from!(self(), moderator, "presenter_answer", %{
+    %{metas: [%{name: name}]} = Presence.get_by_key(socket, email)
```
Rados13, line 223: Couldn't this list contain more than one element?

Rados13 (changes requested):
I think we should increase consistency with the use of client or clientEmail even more, as in some places we use some moderatorTopic in others we pass client. In some place, we pass client in other clientEmail, so I would make it more consistent.

assets/js/components/RtcPlayer.tsx
Karolk99: this is not an Array so it won't work

assets/js/components/PresenterStreams.tsx
Karolk99: because you have to pass it when you are listing components

lib/membrane_live_web/channels/event_channel.ex
Rados13: What about this?

lib/membrane_live_web/channels/event_channel.ex
```diff
@@ -163,17 +165,20 @@ defmodule MembraneLiveWeb.EventChannel do
   # and sends message back) -> server (updates presenter in Presence)
   # such design is caused by user Presence that can be updated only with its socket-channel combination
   # (socket parameter in function below)
-  def handle_in("presenter_remove", %{"presenter" => presenter}, socket) do
-    {:ok, _ref} =
-      Presence.update(socket, presenter, fn map -> Map.put(map, :is_presenter, false) end)
+  def handle_in("presenter_remove", %{"email" => email}, socket) do
+    {:ok, _ref} = Presence.update(socket, email, fn map -> Map.put(map, :is_presenter, false) end)
```
Rados13, line 169: Same as below.

Rados13:
CI fails

Rados13 (approved):

PR #61, opened 2022-09-09
LIVE-67-68-69: Create dashboard UI

LVala:


assets/js/components/dashboard/EventField.tsx
```diff
@@ -0,0 +1,57 @@
...
+  };
+
+  return (
+    <div className="EventField">
+      <div className="EventImageContainer">
+        <Image
+          src="https://drscdn.500px.org/photo/224226723/m%3D2048/v2?sig=98f73dda3c8af34384e01d46702feb68c615f7279b58f0d03cee7e7b4acd82fc"
+          alt="Event Photo"
+          className="EventImage"
+        />
```
Rados13, line 40: Shouldn't we remove this image as we don't plan to add the possibility to add images to the database in the near future?

assets/js/utils/dashboardUtils.ts
```diff
@@ -0,0 +1,61 @@
...
+  description: string;
+  start_date: string;
+  presenters: string[];
+};
+
+export type WebinarInfo = {
+  title: string;
+  description: string;
+  presenters: string[];
+  start_date: Date;
```
Rados13, line 14: As this is exported type this key should be written in camelCase.

assets/js/components/event/Header.tsx
```diff
@@ -41,7 +40,10 @@ const Header = ({ client, eventChannel }: HeaderProps) => {
   const [participantsNumber, setParticipantsNumber] = useState<number>(0);
 
   const navigate = useNavigate();
-  const redirectToHomePage = () => navigate("/");
+  const redirectToHomePage = () => {
+    // TODO
```
Rados13, line 44: What about this `TODO`?

assets/js/components/dashboard/WelcomePanel.tsx
```diff
@@ -0,0 +1,46 @@
...
+        )}
+      </div>
+      <div className="UserContainer">
+        {picture ? (
+          <img src={picture} className="DashboardUserIcon" />
+        ) : (
+          <User1 className="DashboardUserIcon" />
+        )}
+        <div className="DashboardUserName">{name}</div>
+      </div>
```
Rados13, line 41: Aren't this div the same as the div about user information in `Header.tsx`? Could we move it to seperate component?

assets/js/components/dashboard/SideDashboardPanel.tsx
```diff
@@ -0,0 +1,70 @@
...
+    const isActive = currentEvents == text;
+
+    return (
+      <button
+        className="DashboardPanelButton"
+        onClick={onClick}
+        // a very inconvinient way to add hover effect to the icon
+        // as it seems to be impossible in css
+        onMouseEnter={() => setHoverColor(transitionFontColor)}
+        onMouseLeave={() => setHoverColor(bgFontColor)}
```
Rados13, line 31: Did you ask someone more proficient in frontend what is a better way to do this?

assets/js/components/dashboard/SideDashboardPanel.tsx
```diff
@@ -0,0 +1,70 @@
...
+      <div className="DashboardPanelItems PanelFooter">
+        {getButton(QuestionCircle, "Help", () => {
+          // TODO
+        })}
+        {getButton(Logout, "Logout", () => {
+          localStorage.clear();
+          navigate("/auth", { replace: true });
+          // TODO probably some kind of message
+          // needs to be sent to the server to kick the user
+          // from webinars that he's participating in
```
Rados13, line 63: Yup this will be needed

Rados13 (changes requested):
There also many TODO. I think it would be good to add to each TODO whether it will be addressed in this PR or in another.

assets/js/utils/dashboardUtils.ts
```diff
@@ -0,0 +1,61 @@
...
+  eventForm: EventForm,
+  setLink: React.Dispatch<React.SetStateAction<string>>
+): Promise<void> => {
+  axiosWithInterceptor
+    .post("resources/webinars", { webinar: eventForm })
+    .then((response) => {
+      setLink(response.data.link);
+    })
+    .catch((error) => {
+      console.log(error);
```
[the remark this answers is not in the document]
LVala: Fixed.

assets/js/components/dashboard/EventField.tsx
LVala: It would break layout quite severely, i would think about putting there some image with Membrane logo.

assets/js/utils/dashboardUtils.ts
LVala: Fixed (kind of).

assets/js/components/event/Header.tsx
LVala: Implemented.

assets/js/components/dashboard/WelcomePanel.tsx
LVala: Fixed.

assets/js/components/dashboard/SideDashboardPanel.tsx
LVala: Fixed.

assets/js/components/dashboard/SideDashboardPanel.tsx
LVala: We talked about that on the meeting.

assets/js/utils/headerUtils.ts
```diff
@@ -1,31 +1,42 @@
...
-    .then((response) => {
-      const start_date = response.data.webinar.start_date.replace("T", " ");
-      setEventInfo({ ...response.data.webinar, start_date: start_date });
+    .then((response: { data: { webinar: OriginalEventInfo } }) => {
+      const newDate = new Date();
+      newDate.setTime(Date.parse(response.data.webinar.start_date));
+      const newElem: any = { ...response.data.webinar, startDate: newDate };
+      delete newElem.start_date;
+      setEventInfo(newElem as EventInfo);
     })
```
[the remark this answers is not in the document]
LVala: Fixed.

maciejewskakatarzyna:
1. From AC in Jira: 
On the “All events” page, there are two categories of events that are displayed:
- My events - events that have been created by the user
- All events - all events that have been created by users are displayed chronologically (latest →  oldest)
We have "Upcoming events" and "Past events" which is the same as in design but I don't know which version was finally chosen
2. There is also sentence in Jira that the search bar is disabled - it works now, which is not a problem of course but again, I'm not sure what idea stands behind disabling it.
3. Active icons on sidebar should have color #BFCCF8 but now there are white.
4. Active accent should be wider (4px not 2px) because now it looks like some error with render, not as active accent.
5. HiText should be 500 weight (it's 600).
6. Font-family in the majority of components is system-ui not Roc Grotesk / Roboto.
7. Search input and Create event button should have the same height (56px) now one is 50px and the second 48px.
8. What about images? Maybe we should leave some mock images and in future just replace it with real photos?
9. Event presenters should have font-size 18px (line-height 28px) and font-weight 500 not 400.

I am not able to check the dashboard any more because the sandbox page is dead...

LVala:
TODO:
- Take a look at the error mentioned by @maciejewskakatarzyna
- Fix problems with font family

maciejewskakatarzyna:
I think that there should be margin after the last event on the list, because now it is hidden after shadow (screen).
Screenshot 2022-09-20 at 11 27 57

assets/js/pages/Event.tsx
```diff
@@ -68,6 +59,8 @@ const Event = () => {
   }, [eventChannel, privateChannel]);
 
   useEffect(() => {
+    document.title = `${pageTitlePrefix} | ${"a"}`;
```
Rados13, line 62: Why do we add only `a`?

Rados13 (approved):

assets/js/pages/Event.tsx
LVala: Fixed, it was unnecessary.

PR #63, opened 2022-09-12
fix staging and sandbox worklofs

blazpie:


pkrucz00:
The GOOGLE_CLIENT_ID seems to be valid and the request to the google auth endpoint is returning 200. Unfortunately the problem with `your refresh token has expired` comes up again and again due to unknown reasons. I'll see what have happened

pkrucz00:
> 

It looks like the google client id is still not available in runtime. The problem is caused by the lack of an env variable on the backend. The program compares the incoming `aud` claim (which is set correctly) to the nil value from the `MembraneLive.get_env` function

```
iex(2)> MembraneLive.get_env!(:client_id)
nil
iex(3)> MembraneLive.get_env!(:custom_secret)
"secret"
```

Rados13:
> > 
> 
> It looks like the google client id is still not available in runtime. The problem is caused by the lack of an env variable on the backend. The program compares the incoming `aud` claim (which is set correctly) to the nil value from the `MembraneLive.get_env` function
> 
> ```
> iex(2)> MembraneLive.get_env!(:client_id)
> nil
> iex(3)> MembraneLive.get_env!(:custom_secret)
> "secret"
> ```

The problem was that the variable `GOOGLE_CLIENT_ID` wasn't set in `/etc/environment` on the machine 😞 .

blazpie:
Good point @Rados13 Previously env was set with CI inside docker containers, but as we now switched to manual start and we're not using docker to run main application on sandbox, all env variables need to be manually set too

pkrucz00 (approved):

Rados13 (approved):

PR #64, opened 2022-09-13
Remove camera from canvas when disabled

Karolk99:


assets/js/utils/rtcUtils.ts
```diff
@@ -243,11 +251,14 @@ const addOrReplaceTrack = (
 ) => {
   if (!presenterStreams[client.email]) presenterStreams[client.email] = new MediaStream();
   const curTrack = findTrackByType(client, track.kind as SourceType);
-  if (curTrack) {
+
+  if (curTrack && curTrack.id !== track.id && curTrack.id !== mergedScreenRef.screenTrack?.id) {
```
Rados13, line 255: Why do we need these additional conditions in this `if`? Isn't it a bug that we pass to function `addOrReplaceTrack` the track which was already added?

Karolk99: this is not a bug. When we disable user camera we change webrtc track from merged canvas to plain screen share. Then when we do opposite (unable camera) we have to add canvas instead of screen share and canvas has ref to screen share track so we don't want to stop it we just want to remove it from presenterStream object.

Rados13 (approved):

PR #69, opened 2022-09-15
LIVE-130: Participants other than moderator are not authorized to recieve webinar information

LVala:


lib/membrane_live_web/controllers/webinar_controller.ex
```diff
@@ -44,7 +44,7 @@ defmodule MembraneLiveWeb.WebinarController do
          callback
        ) do
     with {:ok, webinar} <- Webinars.get_webinar(uuid),
-         {:ok, webinar_db} <- is_user_authorized(webinar, user_id) do
+         {:ok, webinar_db} <- is_user_authorized(webinar, user_id, callback) do
```
Rados13, line 47: I think we should compare here if this callback variable is `show/2`.

```suggestion
is_show_callback? <- callback == (&show_callback/2)
{:ok, webinar_db} <- is_user_authorized(webinar, user_id, is_show_callback?) do
```

lib/membrane_live_web/controllers/webinar_controller.ex
```diff
@@ -74,8 +74,11 @@ defmodule MembraneLiveWeb.WebinarController do
...
-  defp is_user_authorized(webinar, jwt_user_uuid) when jwt_user_uuid == webinar.moderator_id,
-    do: {:ok, webinar}
+  defp is_user_authorized(webinar, jwt_user_uuid, _callback)
+       when jwt_user_uuid == webinar.moderator_id,
+       do: {:ok, webinar}
 
-  defp is_user_authorized(_webinar, _jwt_user_uuid), do: {:error, :forbidden}
+  defp is_user_authorized(webinar, _jwt_user_uuid, callback) do
+    if callback == (&show_callback/2), do: {:ok, webinar}, else: {:error, :forbidden}
+  end
```
Rados13, line 83: And because change mentioned above it could look like this:
```suggestion
  defp is_user_authorized(webinar, jwt_user_uuid, true), do {:ok, webinar}
  defp is_user_authorized(webinar, jwt_user_uuid, _is_show_callback?)
       when jwt_user_uuid == webinar.moderator_id,
       do: {:ok, webinar}

  defp is_user_authorized(webinar, _jwt_user_uuid, _is_show_callback?), do: {:error, :forbidden}
```

Rados13 (changes requested):

Rados13 (approved):

PR #70, opened 2022-09-16
Live 70 form modal css

pkrucz00:
Adding modal styled with css.


assets/css/main.css
```diff
@@ -14,6 +15,7 @@
   --bg-light-color-3: #bfccf8;
   --bg-light-color-4: #7089db;
   --button-red: #c32222;
+  --dark-blue: #001a72;
```
LVala, line 18: We already have variable for that color, it's `--font-dark-color`

assets/css/main.css
```diff
@@ -23,3 +25,8 @@ html {
   height: 100vh;
   overflow: hidden;
 }
+
```
LVala, line 28: Why is the style for `.FlexContainer` in `main.css` and not in `eventform.css`?

assets/js/components/dashboard/EventForm.tsx
```diff
@@ -0,0 +1,162 @@
...
+  return (
+    <div className="EventFormFieldDiv">
+      <label className="EventFormFieldLabel">Choose presenters (optional)</label>
+      <div className="EventFormFieldInput EventFormFieldWithButtonDiv">
+        <input
+          type="search"
+          placeholder="Type name of presenters to add"
+          onChange={(e) => changeElement(e, setInputPresenter)}
+        />
+        <GenericButton icon={<Search />} onClick={handleAddButton} />
```
LVala, line 104: As I mentioned while chatting, I think it's better to use some kind of "plus" icon instead of `Search`.

assets/css/main.css
pkrucz00: I though it may be generalised since we se it often on frontend.

assets/js/components/dashboard/EventForm.tsx
pkrucz00: Right, I'll change that

assets/js/utils/formApi.ts
```diff
@@ -0,0 +1,18 @@
...
+export const sendEventForm = async (eventForm: EventFormType): Promise<void> => {
+  //TODO change it to an async request and remove the happy path console log
+  axiosWithInterceptor
+    .post("resources/webinars", { webinar: eventForm })
+    .then((response) => {
+      console.log(response.data.link);
+    })
+    .catch((error) => {
+      console.log(error);
+    });
```
Rados13, line 17: Shouldn't it be addressed in this PR?

maciejewskakatarzyna:
1. There is possibility to create event with title " " which throw an error in console but not informed user about it Screenshot 2022-09-16 at 11 08 32
2. There is possibility to add empty string as presenter  Screenshot 2022-09-16 at 11 15 19
3. Search icon suggests that there will be some list of proposed names of presenters but after clicking it, new presenter is added. Maybe we should change the icon OR there should be implemented search logic (but I'm not sure about database of potential presenters).
4. After adding new presenter, the input should be cleared because now when you add new user, data in input are still visible but clicking again on add button adding empty string to the list.
5. There is no possibility to remove presenter from the list (maybe I added someone by mistake and I would like to edit this list).
6. Scrollbar breaks modal layout. Maybe there is possibility to move it outside border? 
 Screenshot 2022-09-16 at 11 22 35 
7. It would be nice if clicking outside the modal would close it (not only close button).
8. There is possibility to add year with 5 numbers (ex. 12345) and it throw error in console (the same as title with spaces only).
9. I know that you added one more extra input and tried to fit its to the current modal size but maybe it will be better to make it a little higher and keep original design margins, sizes and spaces to make it less squeezed?
10. Label's font should be Aktiv Grotesk (now it's system-ui font) and line-height should be 28px (now it's 22.5px).
11. Inputs should have border-radius 40px (the second one 16px) but now it's 25px (15px for the second one).
12. Inputs padding should be 14px 16px now it's 10px.
13. Span with letters left to write should have smaller font-size -> 12px and 16px line-height but now it's 15px.
14. Calendar icon is different than in design. I supposed that it can be found in SWM Icon Pack.
15. Search icon is bigger than in design, it should be 17px x 17px (now it's 24px x 24px)
16. You can write how long event description you want, there is now limitation (it should be 255 according to info in label or 150 according to design - but I understand that you decided it's not enough). 
Screenshot 2022-09-16 at 11 37 40
17. Buttons font should be Aktiv Grotesk (now it's system-ui font), font-size should be 18px (now it's 16px) and line-height should be 24px (now it's 22.5px).
18. Button 'create event' should be 240px x 56px but now it's 231px x 50px and it should have padding 0px 32px (now it's  0px 70px).
19. Maybe it will be better to inform user about required field on the form but make border red and add error message below input than to display toast? In my opinion, user should be inform which field is missing, not informed about all required together.

maciejewskakatarzyna (commented):
I've already left my comments after PR

assets/js/components/dashboard/EventForm.tsx
```diff
@@ -0,0 +1,162 @@
...
+const ModeratorField = () => {
+  return (
+    <div className="EventFormFieldDiv">
+      <label className="EventFormFieldLabel">Choose moderators (optional)</label>
+      <div className="EventFormFieldInput EventFormFieldWithButtonDiv">
+        <input type="search" placeholder="Type name of moderators to add" />
+        <GenericButton
+          icon={<Search />}
+          onClick={() => {
+            console.log("I'm working but please, implement me on backend!");
```
Rados13, line 124: Shouldn't be addressed in this PR?

Rados13 (changes requested):

assets/js/types.ts
```diff
@@ -22,7 +22,7 @@ export type LocationState = {
   pathToReturnTo: string;
 };
 
-export type EventForm = {
+export type EventFormType = {
   title: string;
   description: string;
   start_date: string;
```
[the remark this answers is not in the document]
pkrucz00: It's because the backend expects snake case json variables. It should be changed in the future.

assets/css/main.css
pkrucz00: Changed that

assets/js/components/dashboard/EventForm.tsx
pkrucz00: Changed to a plus

assets/js/utils/formApi.ts
```diff
@@ -1,18 +1,22 @@
...
+
+export const sendEventForm = async (
+  modalType: ModalFormType,
+  eventForm: EventFormType,
+  uuid: string = ""
+): Promise<any> => {
+  const endpoint = "resources/webinars/" + uuid;
+  const method = methodMap[modalType];
+
+  return method(endpoint, { webinar: eventForm });
```
Rados13, line 21: We have a function with the same name in `dashboardUtils.ts`, shouldn't we delete this second function?

Rados13 (commented):
Shouldn't we also add a button somewhere with the possibility of updating the webinar?

Rados13 (commented):
When updating a webinar shouldn't old data load to modal fields?

assets/js/components/dashboard/EventForm.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+type EventFormProps = {
+  setParentInput: (input: EventFormType) => void;
+  defaultInput?: EventInfo;
+};
+
+const EventForm = ({ setParentInput, defaultInput }: EventFormProps) => {
+  const infoToEventForm = (defaultInput: EventInfo): EventFormType => {
+    const convertDateToString = (date: Date) => {
+      const offset = date.getTimezoneOffset() * 60 * 1000;
```
Rados13, line 144: Could you remove magic numbers?

assets/js/components/dashboard/EventForm.tsx
```diff
@@ -0,0 +1,181 @@
...
+      const dateWithOffset = new Date(date.getTime() - offset);
+      return dateWithOffset.toISOString().replace("Z", "");
+    };
+
+    const start_date = convertDateToString(defaultInput.startDate);
+    console.log(start_date);
```
Rados13, line 150: Do we need this `console.log`?

Rados13 (commented):
In addition, it is now impossible to change the Presenter if it was already added.

assets/js/types.ts
```diff
@@ -22,7 +22,7 @@ export type LocationState = {
   pathToReturnTo: string;
 };
 
-export type EventForm = {
+export type EventFormType = {
```
LVala, line 25: By convention in our project, type names does not have `Type` suffix.

assets/js/types.ts
```diff
@@ -45,6 +45,8 @@ export type EventInfo = {
   uuid: string;
 };
 
+export type ModalFormType = "create" | "update";
```
LVala, line 48: Same as above.

assets/js/utils/dashboardUtils.ts
```diff
@@ -28,12 +41,18 @@ export const getWebinarsInfo = async (
...
     })
     .catch((error) => {
       console.log(error);
-      getErrorToast(toast, "Webinar informations could not be obtained...");
+      getErrorToast(toast, "The webinar information could not be obtained.");
     });
 };
+
+const convertDateToString = (date: Date) => {
+  const offset = date.getTimezoneOffset() * 60 * 1000;
```
Rados13, line 55: I would still use `MILLISECONDS_IN_MINUTE` constant.

Rados13 (approved):

LVala (approved):

PR #73, opened 2022-10-03
Update dependencies

Karolk99:
This update is needed for rtc_engine to work correctly in all the scenarios.


Rados13 (commented):
This doesn't compile on the newest version of `http_adaptive_streaming`. Probably because of changes in API due to implementing LL HLS.

Rados13 (changes requested):
I think that tests on the backend for recordings endpoints will be useful.

Rados13 (changes requested):
Also, there are merge conflicts to solve.

Rados13:
Create new PR with proper name

PR #74, opened 2022-10-05
Fix bug with wrong keys

Karolk99:


Rados13 (approved):
The only issue is that the linter fails

PR #79, opened 2022-10-12
Live 146 authenticated user can send request to host to join the stage

Rados13:


assets/js/components/event/ParticipantsList.tsx
```diff
@@ -61,26 +95,36 @@ const Participant = ({ client, participant, eventChannel }: ParticipantProps) =>
...
         className="InfoTooltip"
       >
         {icon}
       </Tooltip>
       <p className="ParticipantText">{participant.name}</p>
+      {participant.requestPresenting && (
+        <Tooltip borderRadius="25px" fontSize={"1.3rem"} className="InfoTooltip">
+          <QuestionCircle className="ParticipantIcon" />
+        </Tooltip>
+      )}
```
LVala, line 117: I think it would be nice to add `label` to the Tooltip with something like "This user is asking to become presenter". Now I'm not sure what does the "?" icon means.

LVala (changes requested):

assets/js/components/event/ParticipantsList.tsx
Rados13: I can't find in raising hand icon, so I decided to go with question mark. This way user can signal that he has some questions and want to say it publicly.

LVala: I don't say that the icon is bad, but that you should add `label` prop to the `Tooltip` with some meaningful message, what's the point of using `Tooltip` without the `label`?

lib/membrane_live_web/channels/event_channel.ex
```diff
@@ -248,4 +209,88 @@ defmodule MembraneLiveWeb.EventChannel do
...
+    do: add_to_list_in_ets(:presenting_requests, email, id)
+
+  defp remove_from_list_in_ets(ets_key, email, id) do
+    [{_key, presenters}] = :ets.lookup(ets_key, id)
+    :ets.insert(ets_key, {id, MapSet.delete(presenters, email)})
+  end
+
+  defp add_to_list_in_ets(ets_key, email, id) do
+    [{_key, presenters}] = :ets.lookup(ets_key, id)
+    :ets.insert(ets_key, {id, MapSet.put(presenters, email)})
```
pkrucz00, line 294: `event_channel.exs` has become quite big with almost 300 lines. Wouldn't it be nice to delgate some functions to i.e. `event_channel_helpers.ex`?

assets/js/components/event/ParticipantsList.tsx
```diff
@@ -61,26 +95,36 @@ const Participant = ({ client, participant, eventChannel }: ParticipantProps) =>
...
       {client.isModerator && (
         <ModeratorMenu
           moderatorClient={client}
           eventChannel={eventChannel}
           participant={participant}
         />
       )}
+      {!client.isModerator && role == "Participant" && isClientPariticipant && (
```
pkrucz00, line 125: I'm a little confused. Why do we need to check 3 things: 
 - the client is not a moderator
 - the client's role is "Participant"
 - the clients is a participant (by email)

How do they differ? Shouldn't the last one imply the rest?
Is there a way to reduce it to 2 checks?

assets/js/components/event/ParticipantsList.tsx
```diff
@@ -61,26 +95,36 @@ const Participant = ({ client, participant, eventChannel }: ParticipantProps) =>
...
-    : "Praticipant";
+    : "Participant";
+
+  const isClientPariticipant: Boolean = client.email == participant.email;
 
   return (
     <div className="Participant">
       <Tooltip
-        label={`${role}${client.name == participant.name ? " (You)" : ""}`}
+        label={`${role}${isClientPariticipant ? " (You)" : ""}`}
```
pkrucz00, line 105: Typo

```suggestion
        label={`${role}${isClientParticipant ? " (You)" : ""}`}
```

assets/js/components/event/ParticipantsList.tsx
```diff
@@ -61,26 +95,36 @@ const Participant = ({ client, participant, eventChannel }: ParticipantProps) =>
...
         borderRadius="25px"
         fontSize={"1.3rem"}
         className="InfoTooltip"
       >
         {icon}
       </Tooltip>
       <p className="ParticipantText">{participant.name}</p>
+      {participant.requestPresenting && (
+        <Tooltip borderRadius="25px" fontSize={"1.3rem"} className="InfoTooltip">
```
pkrucz00, line 114: things like `borderRadius` and `fontSize` should be moved to css file

lib/membrane_live_web/channels/event_channel.ex
```diff
@@ -24,21 +24,24 @@ defmodule MembraneLiveWeb.EventChannel do
...
              is_moderator <- Webinars.check_is_user_moderator(uuid, id) do
           {:ok, socket} = if is_presenter, do: join_event_stream(socket), else: {:ok, socket}
 
           {:ok, _ref} =
             Presence.track(socket, email, %{
               name: name,
               is_moderator: is_moderator,
-              is_presenter: is_presenter
+              is_presenter: is_presenter,
+              request_presenting: is_request_presenting
```
pkrucz00, line 44: wouldn't it be better to stick to the convention?

```suggestion
              is_request_presenting: is_request_presenting
```

pkrucz00 (changes requested):

lib/membrane_live_web/channels/event_channel.ex
Rados13: I looked again through this file and the only functions we could move to the separate module are the current private functions. As the logic of these functions isn't shared with another module I don't see much benefit from moving it to a separate module.

assets/js/components/event/ParticipantsList.tsx
Rados13: Ok, it's bad naming. The last one is checking only if a participant object has the same email as the client object. The goal of this condition determines when the menu for asking for presenting is displayed. It makes sense only for myself when I am not a moderator and I have the role of a participant because if I am a presenter I don't have to ask for it.

lib/membrane_live_web/channels/event_channel.ex
```diff
@@ -232,6 +168,31 @@ defmodule MembraneLiveWeb.EventChannel do
...
   end
 
+  def handle_in(
+        "presenting_request",
+        %{"email" => email},
+        socket
+      ) do
+    {:ok, _ref} = Presence.update(socket, email, &%{&1 | request_presenting: true})
+
+    "event:" <> id = socket.topic
```
Karolk99, line 178: it's just a suggestion would it be better to create id in function pattern match?

lib/membrane_live_web/channels/event_channel.ex
```diff
@@ -232,6 +168,31 @@ defmodule MembraneLiveWeb.EventChannel do
...
+    {:noreply, socket}
+  end
+
+  def handle_in(
+        "cancel_presenting_request",
+        %{"email" => email},
+        socket
+      ) do
+    {:ok, _ref} = Presence.update(socket, email, &%{&1 | request_presenting: false})
+    "event:" <> id = socket.topic
```
Karolk99, line 190: same suggestion

assets/js/components/event/ParticipantsList.tsx
pkrucz00: So maybe it can be moved to a separated function? Or the naming could be clearer.

Karolk99 (changes requested):
One more thing, I'm not sure that everyone should see question-marks those should be only displayed when you are a moderator.

assets/js/components/event/ParticipantsList.tsx
```diff
@@ -59,7 +59,7 @@ const ClientParticipantMenu = ({ participant, eventChannel }: ClientParticipantM
     }
   };
 
-  const text = participant.requestPresenting
+  const text = participant.isRequestPresenting
     ? "Stop asking to become presenter"
     : "Ask to become presenter";
```
pkrucz00, line 64:
```suggestion
    ? "Stop asking to become a presenter"
    : "Ask to become a presenter";
```

assets/js/components/event/ParticipantsList.tsx
```diff
@@ -97,21 +97,17 @@ const Participant = ({ client, participant, eventChannel }: ParticipantProps) =>
...
-        className="InfoTooltip"
-      >
+      <Tooltip label={`${role}${isMyself ? " (You)" : ""}`} className="InfoTooltip">
         {icon}
       </Tooltip>
       <p className="ParticipantText">{participant.name}</p>
-      {participant.requestPresenting && (
-        <Tooltip borderRadius="25px" fontSize={"1.3rem"} className="InfoTooltip">
+      {participant.isRequestPresenting && (client.isModerator || isMyself) && (
+        <Tooltip label={"This user is asking to become presenter"} className="InfoTooltip">
```
pkrucz00, line 110:
```suggestion
        <Tooltip label={"This user is asking to become a presenter"} className="InfoTooltip">
```

pkrucz00 (approved):
Only small grammar issues :D

LVala (approved):

Karolk99 (approved):

PR #80, opened 2022-10-12
Live 145 allow anonymous user to join the stream

Karolk99:
Don't worry about types, there is another pull_request with needed type changes.


lib/membrane_live_web/controllers/webinar_controller.ex
```diff
@@ -59,6 +59,23 @@ defmodule MembraneLiveWeb.WebinarController do
...
+         true <- callback == (&show_callback/2) do
+      callback.(conn, Map.put(params, "webinar_db", webinar))
+    else
+      {:error, :no_webinar} ->
+        %{error: :not_found, message: "Webinar with uuid #{uuid} could not be found"}
+
+      false ->
+        %{error: :forbidden, message: "Unauthenticated user does not have access to this action."}
+    end
+  end
```
Rados13, line 77: We should add tests for these edge cases.

Rados13 (changes requested):
There are merge conflicts to solve.

Rados13 (approved):

PR #81, opened 2022-10-14
Add heart animation

Rados13:


assets/css/event/animation.css
```diff
@@ -0,0 +1,36 @@
...
+
+.heart {
+  z-index: 999;
+  animation: heartfade 6s linear;
+  position: absolute;
+}
+
+.heart:before {
+  content: "";
+  background-color: #fc2a62;
```
LVala, line 19: Maybe it would be nice to use `--dark-font-color` to match the application UI (or maybe we don't want to merge this branch to `master` but to some demo branch, then it's wahtever)

assets/js/components/event/HlsPlayer.tsx
```diff
@@ -17,6 +19,9 @@ const HlsPlayer = ({ hlsUrl, presenterName }: HlsPlayerProps) => {
           <div className="HlsTopBar">
             <div className="HlsPresenterName">{presenterName}</div>
           </div>
+          <div className="HlsBottomBar">
+            <AnimationComponent eventChannel={eventChannel} />
```
LVala, line 23: I was wondering if we could move the reaction button to here and make it some kind of icon, similarly to the demo mock UI.

lib/membrane_live_web/channels/event_channel.ex
```diff
@@ -160,6 +160,13 @@ defmodule MembraneLiveWeb.EventChannel do
     {:noreply, socket}
   end
 
+  def handle_in("reaction", _data, socket) do
+    broadcast(socket, "animation", %{})
+    # push(socket, "animation", %{})
```
LVala, line 165: Are these commented lines necessary?

pkrucz00 (commented):
Maybe the heart emojis could be white? To be closer to the demo

assets/js/components/event/HlsPlayer.tsx
Rados13: We could, but I am not sure how this will compose with the menu of hls-player. I think we could add this when we will change the stream player. By the way for this future task, we could use this:
https://github.com/muxinc/media-chrome

assets/js/components/event/HeartAnimation.tsx
```diff
@@ -0,0 +1,106 @@
...
+      bound: xBound,
+      direction: xStart,
+      scale,
+    };
+  };
+
+  let before = Date.now();
+
+  const frame = () => {
+    let current = Date.now();
```
Karolk99, line 55: should be const

assets/js/components/event/HeartAnimation.tsx
```diff
@@ -0,0 +1,106 @@
...
+    var deltaTime = current - before;
```
Karolk99, line 56: why deltaTime is var and not a const?

assets/js/components/event/HeartAnimation.tsx
```diff
@@ -0,0 +1,106 @@
...
+    const interval = setInterval(() => {
+      frame();
+    }, 5);
+    return () => clearInterval(interval);
+  }, [hearts]);
+
+  const onEvent = () => {
+    var start = 1 - Math.round(Math.random()) * 2;
+    var scale = Math.random() * Math.random() * 0.8 + 0.2;
+    var bound = 30 + Math.random() * 200;
```
Karolk99, line 95: same question

Karolk99 (changes requested):

assets/js/components/event/HeartAnimation.tsx
```diff
@@ -0,0 +1,106 @@
...
+  eventChannel: any;
+};
+
+const AnimationComponent = ({ eventChannel }: HeartAnimationProps) => {
+  const [hearts, setHearts] = useState<HeartElement[]>([]);
+
+  const duration = 5000;
+  const speed = 1;
+  const cursorXOffset = 0;
+  const cursorYOffset = -5;
```
Karolk99, line 24: those const should be moved outside of the component.

assets/js/components/event/HeartAnimation.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const onEvent = () => {
+    const start = 1 - Math.round(Math.random()) * 2;
+    const scale = Math.random() * Math.random() * 0.8 + 0.2;
+    const bound = 30 + Math.random() * 200;
+    return generateHeart(200 + cursorXOffset, 10 + cursorYOffset, bound, start, scale);
+  };
+
+  return (
+    <div className={"animatedContainer"}>
+      {hearts.length}: {hearts.map((heart) => heart.elem)}
```
Karolk99, line 101: i think that is not correct ...?

Karolk99 (changes requested):

Karolk99 (approved):

assets/css/event/animation.css
```diff
@@ -0,0 +1,54 @@
...
+  left: 10.5px;
+  transform: rotate(135deg);
+}
+
+.heartButton {
+  width: 100px;
+  height: 100px;
+  /* width: auto; */
+  /* width: 6rem; */
+  /* height: 6rem; */
```
Karolk99, line 43: it should be removed

Karolk99 (approved):

PR #88, opened 2022-10-21
Added multiple moderators.

roznawsk:


lib/membrane_live/webinars.ex
```diff
@@ -23,9 +24,21 @@ defmodule MembraneLive.Webinars do
...
+    attrs =
+      attrs
+      |> Map.put("creator_id", creator_id)
+      |> Map.update!("moderators", fn moderators ->
+        if moderators do
+          creator_email = MembraneLive.Accounts.get_email!(creator_id)
+          if creator_email in moderators, do: moderators, else: [creator_email | moderators]
+        else
+          nil
+        end
```
Rados13, line 37:
```suggestion
    creator_email = MembraneLive.Accounts.get_email!(creator_id)
    attrs =
      attrs
      |> Map.put("creator_id", creator_id)
      |> Map.update!("moderators", fn 
        nil -> nil
        moderators when creator_email in moderators -> moderators
        moderators -> [creator_email | moderators]
      end)
```

assets/js/components/dashboard/EventForm.tsx
```diff
@@ -142,19 +143,43 @@ const PresenterField = ({ inputList, inputSetter }: ListFieldProps) => {
...
 };
 
-const ModeratorField = () => {
+const ModeratorField = ({ inputList, inputSetter }: ListFieldProps) => {
+  const [inputModerator, setInputModerator] = useState<string>("");
+
+  const handleAddButton = () => {
+    inputModerator.trim() && inputSetter([...inputList, inputModerator.trim()]);
+    setInputModerator("");
+  };
```
Rados13, line 152: I think we should verify here if the user passes a proper email. We could use a famous regex for it:
`/^(([^<>()[\]\\.,;:\s@\"]+(\.[^<>()[\]\\.,;:\s@\"]+)*)|(\".+\"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/`

test/membrane_live/webinars_test.exs
```diff
@@ -55,13 +58,15 @@ defmodule MembraneLive.WebinarsTest do
...
         "presenters" => [],
+        "moderators" => [],
         "start_date" => ~N[2022-07-18 10:20:00],
         "title" => "some updated title"
       }
 
       assert {:ok, %Webinar{} = webinar} = Webinars.update_webinar(webinar, update_attrs)
       assert webinar.description == "some updated description"
       assert webinar.presenters == []
+      assert webinar.moderators == []
```
Rados13, line 69: Shouldn't there be always one email, the email of the creator of the event?

test/support/fixtures/webinars_fixtures.ex
```diff
@@ -20,6 +21,9 @@ defmodule MembraneLive.WebinarsFixtures do
     {:ok, webinar} =
       attrs
       |> Enum.into(@default_webinar_attrs)
+      |> Map.update!("moderators", fn moderators ->
+        if user.email in moderators, do: moderators, else: [user.email | moderators]
+      end)
```
Rados13, line 26: Isn't it the same line as in `MembraneLive.Webinars.create_webinar`? Also shouldn't we guarantee that creator's email is always inside the moderator's list?

Rados13 (changes requested):

test/membrane_live/webinars_test.exs
shuntrho: You're right, we should never remove the owner from moderators.

test/support/fixtures/webinars_fixtures.ex
shuntrho: This line was excessive - we are adding the creator to moderators in `create_webinar` already.

pkrucz00 (commented):
There is a text to fix.

Nonetheless, good work!

test/membrane_live_web/controllers/recordings_controller_test.exs
```diff
@@ -26,7 +26,7 @@ defmodule MembraneLiveWeb.RecordingsControllerTest do
 
     test "lists all recording", %{conn: conn, webinars: webinars} do
       conn = get(conn, Routes.recordings_path(conn, :index))
-      assert json_response(conn, 200)["webinars"] == webinars
+      assert Enum.sort(json_response(conn, 200)["webinars"]) == Enum.sort(webinars)
```
pkrucz00, line 29: This still doesn't resolve the issue of failed test check. Sometimes tests don't pass, at least locally.

pkrucz00 (changes requested):

test/membrane_live_web/controllers/recordings_controller_test.exs
shuntrho: Right, I had to add sorting in the other test as well.

pkrucz00 (approved):
Rebase master to resolve conflicts and you're good to go!

Rados13 (approved):
Looks fine only rebase is needed.

shuntrho:
Work has been suspended, as this feature is not needed for now, and changes have to be made in HLS allowing for multiple moderators.

PR #96, opened 2022-11-03
Add removed features to hls endpoint

Karolk99:

PR #97, opened 2022-11-03
Ml 53 adding icon mute

pkrucz00:
Adding icons indicating user is muted/has camera turned off


assets/js/utils/rtcUtils.ts
```diff
@@ -27,34 +31,45 @@ export const mergedScreenRef: MergedScreenRef = {
 
 const sourceIds: { audio: string; video: string } = { audio: "", video: "" };
 
-export const findTrackByType = (client: Client, sourceType: SourceType) => {
-  return presenterStreams[client.email].getTracks().find((elem) => elem.kind == sourceType);
+export const findTrackByType = (client: Client | Presenter, sourceType: SourceType) => {
```
Rados13, line 34: We probably should create an interface for the common part of `Client` and `Presenter`.

assets/js/utils/rtcUtils.ts
```diff
@@ -27,34 +31,45 @@ export const mergedScreenRef: MergedScreenRef = {
...
+  return presenterStreams[client.email]?.getTracks().find((elem) => elem.kind == sourceType);
```
Rados13, line 35: Why do we need `?` here? If removing it leads to some errors that's mean we call findTrackByType for client that doesn't exists so this is a bug.

assets/js/utils/rtcUtils.ts
```diff
@@ -134,13 +150,23 @@ export const connectWebrtc = async (
...
+      onTrackReady: ({ track, peer }: TrackContext) => {
+        if (track != null) {
+          addOrReplaceTrack(peer.metadata, track, playerCallbacks[peer.metadata.email]);
+          updateMicAndCamStatusCallback();
+        }
+      },
+      onTrackUpdated({ track, peer, metadata }: TrackContext) {
+        if (track != null) {
+          track.enabled = metadata.enabled;
+          addOrReplaceTrack(peer.metadata, track, playerCallbacks[peer.metadata.email]);
```
Rados13, line 165: Why do we need these two lines?

Rados13 (commented):
Also after merges in `membrane_rtc_engine` this branch needs updates from #96.

assets/js/utils/rtcUtils.ts
pkrucz00: We need to change the track `enabled` property to the one received in the metadata. The second line was meant to sync the track with the rest of the tracks.
 I'll check whether or not it works properly  without the second line

pkrucz00: I've checked it. The second line is not necessary. I'll delete that. 🧹

assets/js/utils/rtcUtils.ts
pkrucz00: Yeah, that was right. The function `changeIsTrackEnabled` sometimes was called before the webrtc init. I've changed this function instead of adding this `?` sign.
Hope it is fine. 😅

assets/js/utils/rtcUtils.ts
pkrucz00: Done

Rados13:
image
As the code looks good, I see that these changes are a little bit non-deterministic like in the picture. You should look for more bugs like this in PR or ask QA to test this PR and this way you will get more information.

Rados13 (commented):
image
As the code looks good, I see that these changes are a little bit non-deterministic like in the picture. You should look for more bugs like this in PR or ask QA to test this PR and this way you will get more information.

pkrucz00:
@Rados13  that's a valid point. Do you have a scenario or conditions under which this bug occurs? I couldn't recreate this inconsistency.

Rados13:
> @Rados13 that's a valid point. Do you have a scenario or conditions under which this bug occurs? I couldn't recreate this inconsistency.

The first presenter joins the webinar and disables audio and video tracks, then the second presenter joins.

assets/js/components/event/PresenterStreams.tsx
```diff
@@ -12,10 +17,15 @@ const playerCallbacks: { [key: string]: (sourceType: SourceType) => void } = {};
...
 let webrtcConnecting = false;
 
-const includeKey = (storage: Presenter[], key: string): boolean => {
+const doesIncludeKey = (storage: Presenter[], key: string): boolean => {
   return storage.some((e) => e.email === key);
 };
 
+const useRerender = () => {
+  const [value, setValue] = useState(0);
+  return () => setValue(value + 1);
```
Karolk99, line 26: this function is not component specific so it could be moved to some .ts file

assets/js/types.ts
```diff
@@ -1,5 +1,10 @@
 import { useToast } from "@chakra-ui/react";
 
+export interface BasicUserInfo {
```
Karolk99, line 3: I don't understand the purpose of this type. I guess that the idea was to narrow Client arguments to only the one that are used but in this case what is the purpose of the "name" argument?

Karolk99 (changes requested):

assets/js/components/event/PresenterStreams.tsx
pkrucz00: ok, I'll move it to some file in the `utils` directory.

assets/js/types.ts
pkrucz00: It's basically an intersection between Client and Presenter (and possibly other similar structures like Participant) that was requested by Radek in [this comment](https://github.com/membraneframework-labs/membrane_live/pull/97#discussion_r1017504514).
It is used for functions like `isDeviceEnabled` that were specific only for Client but shouldn't be so.
Other alternatives include passing just the email address, but 
1. the `alias` type would only be an alias for a `string`
2. The `name` information can be helpful in future.

pkrucz00: There is also an option of putting all the types like `Client`, `Participant` etc. into one type that is a set sum of them.  AFAIC this is how the solution from Videoroom new frontend

assets/js/utils/rtcUtils.ts
```diff
@@ -140,13 +159,23 @@ export const connectWebrtc = async (
...
+      onTrackReady: ({ track, peer, metadata }: TrackContext) => {
+        if (track != null) {
+          addOrReplaceTrack(peer.metadata, track, playerCallbacks[peer.metadata.email]);
+          if (metadata.enabled) track.enabled = metadata.enabled;
+          updateMicAndCamStatusCallback();
+        }
+      },
+      onTrackUpdated({ track, metadata }: TrackContext) {
+        if (track != null) {
+          if (metadata.enabled) track.enabled = metadata.enabled;
```
Rados13, line 174: Why can't we always set `track.enabled` to `metadata.enabled`?

Rados13 (commented):
Some bugs still occur like this:
image
For me, it looks like the information isn't propagated somewhere.

assets/js/utils/rtcUtils.ts
pkrucz00: Yeah, that's not what I've meant... it should clearly be `'enabled' in metadata`. Changed it

assets/js/types.ts
Rados13: Change the name to User and change types to interfaces, `Participant` and other types should extend User.

Rados13 (dismissed):

Rados13 (commented):
Icons aren't displayed if the presenter joins with the camera and microphone turned off.

assets/js/components/event/RtcPlayer.tsx
```diff
@@ -24,6 +24,14 @@ const RtcPlayer = ({ isMyself, presenter, playerCallbacks, setPresenters }: RtcP
   };
   playerCallbacks[presenter.email] = connectStreams;
 
+  const isSourceDisabled = (sourceType: SourceType) => {
+    const isEnabled = checkTrackIsEnabled(presenter, sourceType);
+    return isEnabled !== undefined && !isEnabled;
```
Rados13, line 29: Isn't it equal to `isEnabled === false`?

assets/js/utils/rtcUtils.ts
```diff
@@ -140,11 +190,18 @@ export const connectWebrtc = async (
...
+          if ("enabled" in metadata) track.enabled = metadata.enabled;
+          updateMicAndCamStatusCallback();
         }
       },
-      onTrackAdded: () => {
-        // do nothing
+      onTrackUpdated({ track, metadata }: TrackContext) {
+        if (track != null) {
+          if ("enabled" in metadata) track.enabled = metadata.enabled;
+          updateMicAndCamStatusCallback();
```
Rados13, line 201: What is the sense of updating the mic and cam status if we didn't change `track.enabled`?

assets/js/utils/rtcUtils.ts
```diff
@@ -140,11 +190,18 @@ export const connectWebrtc = async (
               },
             };
           });
+
+          if ("enabled" in metadata) track.enabled = metadata.enabled;
+          updateMicAndCamStatusCallback();
```
Rados13, line 195: Same as below?

Rados13 (approved):

assets/js/utils/rtcUtils.ts
```diff
@@ -140,11 +190,18 @@ export const connectWebrtc = async (
               },
             };
           });
+
+          if ("enabled" in metadata) track.enabled = metadata.enabled;
```
Karolk99, line 194:
```suggestion
          track.enabled = metadata?.enabled || track.enabled
```

assets/js/utils/channelUtils.ts
```diff
@@ -152,6 +154,11 @@ export const syncPresenters = (
 
 export const getChannelId = (): string => window.location.pathname.split("/")[2];
 
+export const useRerender = () => {
```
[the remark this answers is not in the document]
pkrucz00: Surely, a person clicking 10 times a second would overflow integer in only 685 mln years. It's only around 0.5 % of the age of he Milky Way. Thank you for finding this dangerous vulnerability.

assets/js/components/event/RtcPlayer.tsx
pkrucz00: Yeah, it is. I'll change that

assets/js/utils/rtcUtils.ts
pkrucz00: If nothing changed, his function has no use. I thought there could possibly be a change of `track.enabled` property that is not from webrtc, but there is no such scenario.

assets/js/utils/rtcUtils.ts
```diff
@@ -140,11 +190,18 @@ export const connectWebrtc = async (
...
+          updateMicAndCamStatusCallback();
         }
       },
-      onTrackAdded: () => {
-        // do nothing
+      onTrackUpdated({ track, metadata }: TrackContext) {
+        if (track != null) {
```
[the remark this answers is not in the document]
pkrucz00: ok

assets/js/utils/rtcUtils.ts
pkrucz00: Maybe it is a little faster, but for me it is not that clear. I'll leave it as it is.

Rados13 (approved):

Karolk99 (approved):

PR #98, opened 2022-11-03
ML-176: Fix problems with RTC engine process proper shutdown

LVala:
- fix to problem with rtc_engine process shutdown
- small cleanup

Possibly there's need to take care of `playlist_playable` message functionality as it seems it does not work properly.


Rados13:
> * fix to problem with rtc_engine process shutdown
> * small cleanup
> 
> Possibly there's need to take care of `playlist_playable` message functionality as it seems it does not work properly.

Some changes with `playlist_playable` are made in #96.

LVala:
> > * fix to problem with rtc_engine process shutdown
> > * small cleanup
> > 
> > Possibly there's need to take care of `playlist_playable` message functionality as it seems it does not work properly.
> 
> Some changes with `playlist_playable` are made in #96.

I've already taken care of it.

lib/membrane_live/event.ex
```diff
@@ -230,6 +230,9 @@ defmodule MembraneLive.Event do
 
       pid == state.moderator_pid ->
         state = %{state | moderator_pid: nil}
+        {:ok, state} = handle_peer_left(state, pid)
+
+        Engine.remove_peer(state.rtc_engine, pid)
```
Rados13, line 235: Hmmm I am not sure about this because, `Engine.remove_peer` sends `Message.PeerLeft`, and in the handler for this message, we call `handle_peer_left again`

LVala: Right, I will look into that.

Rados13 (approved):

PR #103, opened 2022-11-10
Ml 29 presenter preview

shuntrho:


assets/js/components/event/ControlPanel.tsx
```diff
@@ -193,17 +193,27 @@ const ControlPanel = ({
...
+    if (sources != null) {
+      setSources(sources);
+
+      const audio = getCurrentDeviceName(client, "audio");
+      if (audio === undefined)
+        setSourceById(client, sources.audio[0].deviceId, "audio", playerCallback);
+
+      const video = getCurrentDeviceName(client, "video");
+      if (video === undefined)
+        setSourceById(client, sources.video[0].deviceId, "video", playerCallback);
```
Rados13, line 207: I would also check if change like this works.
```suggestion
     const prepareDevice = (kind: "audio" | "video") => {
        const deviceName = getCurrentDeviceName(client, kind);
        if (deviceName === undefined)
          setSourceById(client, sources[kind][0].deviceId, kind, playerCallback);
      }
      prepareDevice("audio");
      prepareDevice("video")
```

Rados13 (changes requested):
I found a few issues, but as a whole PR looks good. Nice usage of an existing codebase, I thought it would require more changes. 
During testing PR I found a bug that after the second presenter joins, the main presenter doesn't see him and receives an error in the console 
```
Uncaught TypeError: Cannot read properties of undefined (reading 'connectCallbacks')
    at Object.onTrackReady (rtcUtils.ts:143:43)
    at RTCPeerConnection.<anonymous> (membraneWebRTC.js:383:99)
```

assets/js/components/event/ControlPanel.tsx
```diff
@@ -175,7 +175,7 @@ const useRerender = () => {
 
 type ControlPanelProps = {
   client: Client;
-  webrtc: MembraneWebRTC;
+  webrtc: MembraneWebRTC | null;
```
pkrucz00, line 178: I assume this is necessary to make a preview but this may allow the occurrence of the bug I've mentioned in the main review comment.

assets/js/components/event/PresenterArea.tsx
```diff
@@ -0,0 +1,129 @@
...
+  useEffect(() => {
+    Object.values(presenters).forEach((presenter) => {
+      if (presenter.status == "connecting" && presenter.connectCallbacks.length > 0) {
+        const playerCallback = playerCallbacks[presenter.email];
+        presenter.connectCallbacks.forEach((callback) => callback(playerCallback));
+        setPresenters({
+          ...presenters,
+          [presenter.email]: { ...presenter, connectCallbacks: [] },
+        });
+      }
```
pkrucz00, line 71: I'd suggest putting this function in the `rtcutils.ts` file

assets/js/components/event/PresenterArea.tsx
```diff
@@ -0,0 +1,129 @@
...
+          playerCallback={playerCallbacks[client.email]}
+          setMode={setMode}
+        />
+      )}
+      {!isClientPresenting && (
+        <button className="StartPresentingButton" onClick={onPresenterReady}>
+          Start presenting
+        </button>
+      )}
+    </div>
```
pkrucz00, line 123: I fill like the presenter preview should additionally  have a title  indicating that this is a preview (like `Presenter Preview` or something like that).
Also: consider moving this part to a different function inside `PresenterArea.tsx` or about moving this part to a separate component

assets/js/utils/rtcUtils.ts
```diff
@@ -1,14 +1,18 @@
...
 
 export type Sources = {
   audio: MediaDeviceInfo[];
   video: MediaDeviceInfo[];
 };
-export type SourceType = "audio" | "video";
+
+export type SourcesInfo = {
+  audio: MediaDeviceInfo | undefined;
+  video: MediaDeviceInfo | undefined;
```
pkrucz00, line 14: Can we avoid using `undefined` here? It may lead to some nasty bugs

pkrucz00 (commented):
When second presenter joins there is often a problem with some components checking if track is enabled before they are connected with the webrtc. This result in:
   - error on the joining presenter site: 
```Uncaught TypeError: Cannot read properties of undefined (reading 'getTracks')
    at findTrackByType (rtcUtils.ts:35:38)
    at checkTrackIsEnabled (rtcUtils.ts:58:9)
    at getMuteButton (ControlPanel.tsx:243:11)
(...)
```
  - no new presenter on the other presenter site
  - error on backend:
 ```
  * 1st argument: not a pid

    :erlang.node(nil)
    (membrane_live 0.1.0) lib/membrane_live/event.ex:131: MembraneLive.Event.handle_info/2
    (stdlib 3.17.2) gen_server.erl:695: :gen_server.try_dispatch/4
    (stdlib 3.17.2) gen_server.erl:771: :gen_server.handle_msg/6
    (stdlib 3.17.2) proc_lib.erl:226: :proc_lib.init_p_do_apply/3
Last message: %Membrane.RTC.Engine.Message.NewPeer{peer: %Membrane.RTC.Engine.Peer{id: "c62267cd-56cf-4462-a6f5-ee7b6e23db1b", metadata: %{"email" => "email@swmansion.com", "isAuthenticated" => true, "isModerator" => false, "name" => "Forname Surname"}}, rtc_engine: #PID<0.1077.0>}
```

File: assets/js/components/event/ControlPanel.tsx (hunk 2 above)
Comment (shuntrho) [reply assets/js/components/event/ControlPanel.tsx:178]: We can add Tracks before we are connected to webrtc, so the potential bug is not really dependent on this.

File: assets/js/utils/rtcUtils.ts [hunk 6]
```diff
@@ -17,7 +21,7 @@ export type MergedScreenRef = {
   refreshId: NodeJS.Timeout | undefined;
 };
 
-export const presenterStreams: { [key: string]: MediaStream } = {};
+export const presenterArea: { [key: string]: MediaStream } = {};
```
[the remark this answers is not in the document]
shuntrho: Yeah, since this component contains preview, which isn't connected to RtcEngine, and no data is received from it I wouldn't say there are any *Streams* there.

assets/js/components/event/PresenterArea.tsx
shuntrho: I moved the RtcPlayer into a function, the preview indication is up for a debate.

assets/js/components/event/PresenterArea.tsx
```diff
@@ -24,87 +30,86 @@ const PresenterArea = ({ client, eventChannel, mode, setMode }: PresenterAreaPro
   const [isControlPanelAvailable, setIsControlPanelAvailable] = useState(false);
   const [isClientPresenting, setIsClientPresenting] = useState(false);
 
+  const onPresenterReady = () => {
+    setIsClientPresenting(true);
+
+    eventChannel &&
+      eventChannel.push("presenter_ready", {
+        email: client.email,
+      });
```
Rados13, line 39:
```suggestion
      eventChannel?.push("presenter_ready", {
        email: client.email,
      });
```

assets/js/components/event/PresenterArea.tsx
```diff
@@ -24,87 +30,86 @@ const PresenterArea = ({ client, eventChannel, mode, setMode }: PresenterAreaPro
...
+  };
+
+  const getCurrentPresenter = () => {
+    return (
+      Object.values(presenters).find((presenter) => presenter.email == client.email) || {
+        name: client.name,
+        email: client.email,
+        status: "idle",
+        connectCallbacks: [],
+      }
```
Rados13, line 49: I would move this object maybe to a constant like `idleClient`.

assets/js/components/event/PresenterArea.tsx
```diff
@@ -24,87 +30,86 @@ const PresenterArea = ({ client, eventChannel, mode, setMode }: PresenterAreaPro
...
       webrtcConnecting = true;
-      connectWebrtc(eventChannel, client, presenters, setPresenters).then((value) => {
+      connectWebrtc(eventChannel, client, setPresenters).then((value) => {
         webrtc = value;
         setIsControlPanelAvailable(true);
         webrtcConnecting = false;
       });
-    } else if (webrtc != null && !isClientPresenter) {
+    } else if (webrtc != null && clientShouldDisconnect) {
       leaveWebrtc(webrtc, client, eventChannel);
```
Rados13, line 74:
```suggestion
    } else if (clientShouldDisconnect) {
      leaveWebrtc(webrtc!, client, eventChannel);
```

assets/js/utils/rtcUtils.ts
```diff
@@ -184,6 +181,24 @@ export const connectWebrtc = async (
...
+      .filter(
+        (presenter) => presenter.status == "connecting" && presenter.connectCallbacks.length > 0
+      )
+      .forEach((presenter) => {
+        const playerCallback = playerCallbacks[presenter.email];
+        presenter.connectCallbacks.forEach((callback) => callback(playerCallback));
+        presenters = { ...presenters, [presenter.email]: { ...presenter, connectCallbacks: [] } };
+      });
+    return presenters;
+  });
```
Rados13, line 199: Could we move this to some function like `doPresenterCallback` or something like this?

Rados13 (commented):
I will check it also tomorrow.

Rados13 (changes requested):
There is also a bug that crashes the frontend. Steps for reproduction:
- join webrtc as presenter
- stop being a presenter
- become presenter again

assets/js/components/event/PresenterArea.tsx
shuntrho: Can't do that unfortunately, linter **screms**.

Karolk99 (commented):
I looked into this code and I think that we can get rid of creating `RtcPlayers` based on list of presenters. We can spawn `RtcPlayers` by mapping `presentersStreams` which has all available tracks. For that purpose `presenterStreams` would have to become a React state.

shuntrho:
> I looked into this code and I think that we can get rid of creating `RtcPlayers` based on list of presenters. We can spawn `RtcPlayers` by mapping `presentersStreams` which has all available tracks. For that purpose `presenterStreams` would have to become a React state.

Do we want to do this in this PR, or maybe in a new one. Because, although quite ugly the current solution works fine? @Rados13

Rados13 (approved):

Rados13 (approved):

PR #108, opened 2022-11-23
Fix presenter reentering event

shuntrho:
When the presenter enters the event, when no peers are present it caused an error, because the event wasn't started. Now the event is started when moderator or presenter joins.

### DIFF
--- a/lib/membrane_live/webinars.ex
+++ b/lib/membrane_live/webinars.ex
@@ -79,8 +79,8 @@ defmodule MembraneLive.Webinars do
     "/event/#{webinar.uuid}"
   end
 
-  @spec check_is_user_moderator(binary(), binary()) :: boolean()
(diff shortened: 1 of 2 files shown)


Karolk99 (changes requested):
Reentering is fixed but right now until sessionStorage is automatically cleaned by browser, previous presenter will always enter event as presenter. I don't think that is the effect that we wanted. @Rados13 ?

Rados13:
> Reentering is fixed but right now until sessionStorage is automatically cleaned by browser, previous presenter will always enter event as presenter. I don't think that is the effect that we wanted. @Rados13 ?

I would talk with Krystian about it, as I don't have any preferences.

shuntrho:
We changed the logic of starting and shutting down event in #115, so I am closing this PR and removing this branch @Rados13.

PR #111, opened 2022-11-29
Adjusting backarrows

pkrucz00:


LVala (approved):
I noticed the tooltip on icons next to participant names in participant list got broken, maybe you could fix by the way.

LVala (approved):

PR #115, opened 2022-12-02
Ml 140 last person shuts webinar

shuntrho:
For testing purposes, the timeout for the notification and closing the webinar are set to 10 seconds.


Rados13 (commented):
Also, I think we could add tests for timer and ending an event.

Rados13 (approved):
One last thing, more UI/UX is that a popup informing about closing an event probably shouldn't wait for action from the user if the second user comes (the popup should automatically close if the second user comes). But besides that, it can go to QA.

Rados13 (dismissed):
Looks good it can goes to QA

Rados13 (approved):

PR #116, opened 2022-12-05
Ml 198 fix presenter preview

shuntrho:


Rados13 (approved):
After the second try, everything is working properly 😅, sorry for the false information.

PR #117, opened 2022-12-06
Fix typechecking

LVala:


assets/js/services/jwtApi.ts
```diff
@@ -1,9 +1,14 @@
...
 
-export const addJwtToHeader = (config) => {
-  config.headers.Authorization = getAuthBearer();
-  config.headers.RefreshToken = storageGetRefreshToken();
+export const addJwtToHeader = (config: AxiosRequestConfig<unknown>) => {
+  if (config.headers) {
+    config.headers.Authorization = getAuthBearer();
+    const refreshToken = storageGetRefreshToken();
+    if (refreshToken) config.headers.RefreshToken = refreshToken;
+    else console.log("Jwt refresh token is not defined");
```
Rados13, line 10: I think that should be `console.error` if we don't have a refresh token but we have authToken.

assets/js/utils/rtcUtils.ts
```diff
@@ -24,7 +24,7 @@ export const mergedScreenRef: MergedScreenRef = {
   refreshId: undefined,
 };
 
-const sourceIds: { audio: string; video: string } = { audio: "", video: "" };
+const sourceIds: { audio: string; video: string; [key: string]: string } = { audio: "", video: "" };
```
Rados13, line 27: Why do we need this change?

Rados13 (approved):

assets/js/utils/rtcUtils.ts
LVala: Unfortunately, the `MediaStreamTrack.kind`, used later in the file as a key to the object, is of type `string` even tho [it can only take values `video` or `audio`](https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack/kind), thus this change.

Rados13 (approved):

PR #122, opened 2022-12-14
Adding calculation of main elements with the web browser bar

pkrucz00:


pkrucz00:
There is a new feature `dvw`. I need to check that

assets/js/components/Router.tsx
```diff
@@ -1,10 +1,19 @@
...
 import Recording from "../pages/Recording";
 
 const Router = () => {
+  useEffect(() => {
+    const changeCssHeightVariable = () =>
+      window.document.documentElement.style.setProperty("--window-inner-height", `${window.innerHeight}px`);
+    changeCssHeightVariable();
+    window.addEventListener("resize", changeCssHeightVariable);
+
+    return window.removeEventListener("resize", changeCssHeightVariable);
```
Rados13, line 14: Shouldn't it be a function?

pkrucz00: Should be. Changed that

pkrucz00:
Pushed `ML-157-presenter-on-android` branch here by accident. Reopening

PR #124, opened 2022-12-15
Ml 203 dvw

pkrucz00:
Dynamic view port is a new feature that has been recently introduced to the web browsers. Currently it is used in about 60% of web browsers ([source](https://caniuse.com/viewport-unit-variants)) This solution must be taken under consideration since it is elegant and simple, but the [parallel PR](https://github.com/membraneframework-labs/membrane_live/pull/122) seems to work on almost every web browser


pkrucz00:
Since the site is unusable on mobiles that do not support `dvw` this branch is closed

PR #126, opened 2022-12-20
ML-182: Chat persistance

LVala:


lib/membrane_live/chats.ex
```diff
@@ -0,0 +1,64 @@
...
+      }
+    )
+    |> Repo.all()
+  end
+
+  def add_chat_message(event_id, user_name, user_email, is_auth, content, time_offset) do
+    case is_auth do
+      true -> add_authenticated_chat_message(event_id, user_email, content, time_offset)
+      false -> add_anonnymous_chat_message(event_id, user_email, user_name, content, time_offset)
+    end
```
Rados13, line 34: Isn't it an if?

lib/membrane_live_web/channels/event_channel.ex
```diff
@@ -199,12 +200,44 @@ defmodule MembraneLiveWeb.EventChannel do
...
+          end
+        else
+          case :ets.lookup(:client_start_timestamps, id) do
+            [{_key, client_start_timestamp}] ->
+              System.monotonic_time(:millisecond) - client_start_timestamp
+
+            [] ->
+              0
+          end
+        end
```
Rados13, line 230:
```suggestion
      key = if is_auth and is_presenter, do: :start_timestamps, else: :client_start_timestamps

      new_offset =
        cond do
          case :ets.lookup(key, id) do
            [{_key, start_timestamp}] ->
              System.monotonic_time(:millisecond) - start_timestamp

            #  TODO handle messages when presenter is in the "start streaming" window
            [] when is_auth and is_presenter ->
              raise "Recieved chat message from presenter when there is no presenters connected to RTC Engine process"

            [] ->
              0
          end
        end
```

lib/membrane_live_web/channels/event_channel.ex
```diff
@@ -199,12 +200,44 @@ defmodule MembraneLiveWeb.EventChannel do
...
+
+      {:ok, is_presenter} = check_if_presenter(email, true, id)
+
+      new_offset =
+        if is_auth and is_presenter do
+          case :ets.lookup(:start_timestamps, id) do
+            [{_key, start_timestamp}] ->
+              System.monotonic_time(:millisecond) - start_timestamp
+
+            #  TODO handle messages when presenter is in the "start streaming" window
```
Rados13, line 218: This TODO will be addressed in this PR or in another?

lib/membrane_live_web/controllers/chat_controller.ex
```diff
@@ -0,0 +1,22 @@
...
+  use MembraneLiveWeb, :controller
+
+  alias MembraneLive.Chats
+
+  action_fallback(MembraneLiveWeb.FallbackController)
+
+  def index(conn, %{"uuid" => id}) do
+    chat_messages =
+      Chats.get_event_chat_messages(id)
+      |> Enum.map(
```
Rados13, line 11:
```suggestion
      id
      |> Chats.get_event_chat_messages()
      |> Enum.map(
```

lib/membrane_live_web/channels/event_channel.ex
LVala: Is the `cond do` block intentional?

Rados13: Nope, it's a typo, it should be the only case, sorry for a mistake 😅

lib/membrane_live_web/channels/event_channel.ex
```diff
@@ -233,33 +243,24 @@ defmodule MembraneLiveWeb.EventChannel do
...
-        if is_auth and is_presenter do
-          case :ets.lookup(:start_timestamps, id) do
-            [{_key, start_timestamp}] ->
-              System.monotonic_time(:millisecond) - start_timestamp
+      key = if is_auth and is_presenter, do: :start_timestamps, else: :client_start_timestamps
 
-            #  TODO handle messages when presenter is in the "start streaming" window
-            [] ->
-              raise "Recieved chat message from presenter when there is no presenters connected to RTC Engine process"
-          end
```
Rados13: So in a situation like this will we simply send messages without delay?

Rados13 (approved):
Looks nice it can go to QA.

lib/membrane_live_web/channels/event_channel.ex
LVala: Before such case would indicate that something is broken, but now one can be a 'presenter' and not stream anything through `hls`, so I believe that this is appropriate.

Rados13 (dismissed):

Rados13 (approved):

PR #128, opened 2022-12-22
Ml 137 products on backend

pkrucz00:
Adding endpoints described [here](https://docs.google.com/document/d/1TWyGnCkSfvZAqCY4oyieBlyWokN8-TaLQ4z3aLVZ3bE/edit?pli=1#bookmark=id.6w2s1mz3wssi).
In short: this feature should enable sending HTTP requests in order to add or delete product from a webinar, list all products (all or by webinar).
Also it changes the database (adds `products` table and `webinars_products`(an intermediate table between `products` and `webinars`) and adds a scripts that populates the `products` table (the `priv/repo/seed.exs` file)


lib/membrane_live/products/product.ex
```diff
@@ -0,0 +1,33 @@
...
+  """
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  @derive {Phoenix.Param, key: :uuid}
+  @primary_key {:uuid, :binary_id, autogenerate: true}
+
+  @type t :: %__MODULE__{
+          name: String.t(),
+          price: String.t(),
```
Rados13, line 13: It's a detail but price, probably shouldn't be a string type.

lib/membrane_live/webinars.ex
```diff
@@ -79,11 +87,18 @@ defmodule MembraneLive.Webinars do
...
+    end
+  end
+
+  @spec check_is_user_moderator!(binary(), binary()) :: boolean()
+  def check_is_user_moderator!(user_uuid, webinar_uuid) do
+    case check_is_user_moderator(user_uuid, webinar_uuid) do
+      :ok -> true
+      {:error, _reason} -> false
     end
   end
```
Rados13, line 103: According to the elixir convention, the function with a bang should raise in case of error, which we don't do here. So instead I think we should rename this function in some way because the possibility of identifying what type of error we receive can be pretty useful.
More description about this convention: https://stackoverflow.com/a/33326895

lib/membrane_live/webinars_products.ex
```diff
@@ -0,0 +1,49 @@
...
+      :ok
+    end
+  end
+
+  @spec product_in_webinar?(Product.t(), Webinar.t()) :: boolean
+  def product_in_webinar?(product, webinar) do
+    case Repo.get_by(WebinarsProducts, webinar_id: webinar.uuid, product_id: product.uuid) do
+      nil -> false
+      _webinar_product -> true
+    end
```
Rados13, line 47:
```suggestion
Repo.get_by(WebinarsProducts, webinar_id: webinar.uuid, product_id: product.uuid) != nil
```

lib/membrane_live_web/helpers/controller_callback_helper.ex
```diff
@@ -4,14 +4,26 @@ defmodule MembraneLiveWeb.Helpers.ControllerCallbackHelper do
 
   alias MembraneLive.Webinars
 
-  def get_with_callback(
+  def get_webinar_and_fire_callback(
+        conn,
+        params,
+        callback,
+        show_callback? \\ false,
+        products? \\ false
```
Rados13, line 12: I think name `include_products?` would be better, but if you have a strong preference for this name it can stay as it is.

priv/repo/seeds.exs
```diff
@@ -9,3 +9,20 @@
...
+
+    IO.puts("Adding products")
+    Products.create_product!(%{name: "Maybeline Ink", price: "21.37", itemUrl: mock_item_url , imageUrl: mock_image_url})
+    Products.create_product!(%{name: "Maybeline Highlighter", price: "10.29", itemUrl: mock_item_url , imageUrl: mock_image_url})
+    Products.create_product!(%{name: "Maybeline Eyelasher", price: "74.39", itemUrl: mock_item_url , imageUrl: mock_image_url})
+    Products.create_product!(%{name: "Maybeline Eye Magic", price: "123.34", itemUrl: mock_item_url , imageUrl: mock_image_url})
+
+  _ ->
+    nil
+end
```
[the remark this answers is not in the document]
pkrucz00: Thanks:)

Rados13 (approved):
There are 3 issues from the previous review, but they are crucial so we can merge them and fix them if they become a problem.

PR #131, opened 2022-12-29
LL-HLS

shuntrho:


lib/membrane_live_web/controllers/hls_controller.ex
```diff
@@ -14,4 +68,98 @@ defmodule MembraneLiveWeb.HLSController do
...
+
+    partial_ets_name = "muxed_segment_#{segment}_#{filename_without_extension}_#{offset}"
+
+    case find_partial_in_ets(partial_ets_name) do
+      [{_key, content}] ->
+        {:ets, content}
+
+      [] ->
+        segment_path = Helpers.hls_output_path(prefix, segment_filename)
+        segment_size = File.stat!(segment_path) |> Map.fetch!(:size)
```
Rados13, line 131:
```suggestion
        segment_size = segment_path |> File.stat!() |> Map.fetch!(:size)
```

Rados13 (commented):
I also think it will be helpful to move some utils function about parsing manifest/playlist to separate modules because I think you share some logic.

Rados13 (approved):
I think that the only thing left regarding ll-hls is to automate the measurement of latency, but we have a separate [task for it ](https://membraneframework.atlassian.net/browse/ML-217?atlOrigin=eyJpIjoiZTU0NTA1YWY0MzgzNGM1ZTk3MWMxNTc4YWVjM2Y2ZTgiLCJwIjoiaiJ9), so this PR can go to QA.

Rados13 (approved):

PR #133, opened 2023-01-02
Fixed presenter joining WebRTC without any tracks.

shuntrho:
Fixed issue with presenter joining without any tracks connected.
That happened when we joined immediately after becoming a presenter. Connecting tracks takes some time and if we joined before that it caused the peer to join without audio or video tracks and made the engine crash.


LVala (approved):
[👍](https://emojipedia.org/thumbs-up/)

pkrucz00 (approved):

pkrucz00:
👍

PR #134, opened 2023-01-03
ML-218: Make chat respond to jumps on playback timeline in recording

LVala:


assets/js/utils/useRecordingChatMessages.ts
```diff
@@ -0,0 +1,65 @@
...
+    const id = window.location.href.split("/").slice(-1)[0];
+    axios
+      .get(`${window.location.origin}/resources/chat/${id}`)
+      .then(({ data: { chats: prevChatMessages } }: { data: { chats: AwaitingMessage[] } }) => {
+        if (!ignore) {
+          chatMessagesStack.current = prevChatMessages.reverse();
+          setIsChatLoaded(true);
+        }
+      })
+      .catch((error) => console.log("Fetching previous chat messages failed: ", error));
```
Rados13, line 57: Probably it should be `console.error`

assets/js/components/event/ChatBox.tsx
```diff
@@ -54,14 +54,16 @@ const MessageBox = ({ message, isMyself }: MessageBoxProps) => {
   };
 
   return (
-    <div className={`MessageBox ${isMyself ? "Own" : "Other"}`} key={message.contents[0]}>
```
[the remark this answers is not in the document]
LVala: I don't know why it was there in the first place. The component is used in a list in line 123 of `ChatBox.tsx`, where key is assigned where it's actually needed.

assets/js/utils/useRecordingChatMessages.ts
```diff
@@ -11,54 +11,55 @@ export const useRecordingChatMessages = (): {
...
-    setChatMessages((prev) => appendToMessages(prev, messagesToAdd, undefined));
+    if (content && message) message.contents.push(content);
+    if (message) prev.push(message);
 
-    if (messagesToAdd.length === 0) {
-      setChatMessages((prev) => {
-        const messagesToRemove: AwaitingMessage[] = [];
+    chatMessagesStack.current.push(...messagesToRemove);
+    return [...prev];
+  }, []);
```
Rados13, line 31: I want to make sure how it works. We have two stacks one with messages already displayed and the second with messages to display. When we are removing messages we simply take displayed messages from the stack and push them to the stack with messages not displayed.
It also looks very similar to `addMessage` function so maybe we could create a generic function for both use cases?

Rados13 (approved):
Besides these small issues, it can go to QA.

assets/js/utils/useRecordingChatMessages.ts
LVala: It works exactly as you explained it. I'm not sure about creating generic function as the logic behind pushing and poping from the shown messages is quite different, but I will try to simplify it a bit more.

Rados13 (approved):

PR #138, opened 2023-01-05
Ml 144 improved animations

pkrucz00:
1. Improved animation of floating heart (centred and visible for presenters).
2. Added confetti animation.

### DIFF
--- a/assets/js/components/event/ModePanel.tsx
+++ b/assets/js/components/event/ModePanel.tsx
@@ -3,10 +3,10 @@ import { Screen } from "react-swm-icon-pack";
 import { syncPresentersNumber } from "../../utils/modePanelUtils";
 import { Channel } from "phoenix";
 import type { Client, Mode } from "../../types/types";
+import { useStateTimeout } from "../../utils/reactUtils";
 
 import "../../../css/event/modepanel.css";
 import "../../../css/event/animation.css";
-import { useStateTimeout } from "../../utils/reactUtils";
 
 export type ModeButtonProps = {
   onClick: () => void;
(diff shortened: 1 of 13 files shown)


lib/membrane_live_web/channels/event_channel.ex
```diff
@@ -219,8 +219,13 @@ defmodule MembraneLiveWeb.EventChannel do
...
-    broadcast(socket, "animation", %{})
+  def handle_in("heart_reaction", _data, socket) do
+    broadcast(socket, "heart_animation", %{})
+    {:noreply, socket}
+  end
+
+  def handle_in("confetti_reaction", _data, socket) do
+    broadcast(socket, "confetti_animation", %{})
     {:noreply, socket}
   end
```
Rados13, line 230: What do you think about a change like this?
```suggestion
  def handle_in(animation_type <> "_reaction", _data, socket) when animation_type in ["confetti", "heart"] do
    broadcast(socket,animation_type  <>"_animation", %{})
    {:noreply, socket}
  end
```

Rados13 (dismissed):
The code looks good it can go to QA

lib/membrane_live_web/channels/event_channel.ex
pkrucz00: Looks great! I'll commit it right away

pkrucz00: Although the code looks good, it does not work. The `<>` operator needs to have an explicit byte_size of all the arguments excluding the last one. Fixed that by switching animation type with `"reaction"` (it's `reaction_heart` now, not `heart_reaction`).

assets/js/components/event/ModePanel.tsx
```diff
@@ -31,23 +32,36 @@ type ModePanelProps = {
...
+    };
+  }, [isHeartClicked]);
+
+  useEffect(() => {
+    const ref = setTimeout(() => setConfettiClicked(false), 2_500);
     return () => {
       clearTimeout(ref);
     };
-  }, [isClicked]);
+  }, [isConfettiClicked]);
```
kamil-stasiak, line 55: You could create custom hook for that. It toggles value for some time.

assets/js/components/event/animations/ConfettiAnimation.tsx
```diff
@@ -0,0 +1,29 @@
...
+    const ref = eventChannel?.on("animation_confetti", () => {
+      setIsRun(true);
+      timeoutRef = setTimeout(() => setIsRun(false), 2_500);
+    });
+
+    return () => {
+      timeoutRef && clearTimeout(timeoutRef);
+      eventChannel?.off("animation_confetti", ref);
+    };
+  }, [eventChannel]);
```
kamil-stasiak, line 24: maybe it is possible to include this logic in previously suggested hook

assets/js/components/event/animations/HeartAnimation.tsx
```diff
@@ -80,23 +80,23 @@ const AnimationComponent = ({ eventChannel }: HeartAnimationProps) => {
...
+    const randomXFactor = Math.random() * 200 - 100;
     const randomYFactor = Math.random() * 20;
 
-    return generateHeart(200 + cursorXOffset + randomXFactor, 10 + cursorYOffset + randomYFactor, bound, start, scale);
+    return generateHeart(cursorXOffset + randomXFactor, 10 + cursorYOffset + randomYFactor, bound, start, scale);
   }, [generateHeart]);
 
   useEffect(() => {
-    const ref = eventChannel?.on("animation", () => {
+    const ref = eventChannel?.on("animation_heart", () => {
```
kamil-stasiak, line 91: every messageId could be constant

kamil-stasiak (approved):

PR #139, opened 2023-01-10
ML-137 Fix ios chrome bottom bar animation

kamil-stasiak:


assets/js/types/types.ts
```diff
@@ -149,4 +149,4 @@ export type Product = {
   imageUrl: string;
 };
 
-export type Card = "hidden" | "hide" | "share" | "chat" | "products";
+export type CardStatus = "hidden" | "share" | "chat" | "products";
```
Rados13, line 152: Shouldn't it be typescript [enum](https://www.typescriptlang.org/docs/handbook/enums.html)?

Rados13 (approved):

PR #149, opened 2023-01-13
Changing axios interceptor

pkrucz00:
After this change the axios interceptor:
 1. Doesn't redirect to \auth thus does not directing to the non existent site
 2. Requires to handle the rejected promise by the developer, passing the responsibility further into the code

### DIFF
--- a/lib/membrane_live/tokens/custom_tokens.ex
+++ b/lib/membrane_live/tokens/custom_tokens.ex
@@ -7,7 +7,7 @@ defmodule MembraneLive.Tokens.AuthToken do
 
   import MembraneLive.Helpers
 
-  @one_day 1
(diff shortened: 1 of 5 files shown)


kamil-stasiak (approved):

lib/membrane_live/tokens/custom_tokens.ex
```diff
@@ -26,7 +26,7 @@ defmodule MembraneLive.Tokens.RefreshToken do
 
   import MembraneLive.Helpers
 
-  @one_week 7 * 24 * 60 * 60
+  @one_week 7
```
kamil-stasiak, line 29: It looks like it should not be changed.

kamil-stasiak (changes requested):

PR #156, opened 2023-01-23
Prevent event from starting once it has been finished.

shuntrho:
Firstly, so far the event (engine) hasn't even been closed when moderator decided to end the webinar.
Then, after engine termination has been added it turns out that a moderator tried to join the event_channel, which caused another engine to start and wipe out the HLS directory, so no recording was available. The newly started engine probably wasn't killed as well.

### DIFF
--- a/lib/membrane_live/event.ex
+++ b/lib/membrane_live/event.ex
@@ -328,23 +328,20 @@ defmodule MembraneLive.Event do
     end
   end
 
-  defp close_webinar(state) do
-    Engine.terminate(state.rtc_engine)
-
-    MembraneLiveWeb.Endpoint.broadcast!("event:" <> state.event_id, "finish_event", %{})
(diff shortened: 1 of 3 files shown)


Karolk99 (dismissed):

lib/membrane_live/event.ex
```diff
@@ -342,6 +342,12 @@ defmodule MembraneLive.Event do
     {:reply, stream_response_message(state), state}
   end
 
+  @impl true
+  def handle_cast(:finish_event, state) do
+    Engine.terminate(state.rtc_engine)
+    {:stop, :normal, state}
```
Rados13, line 348: I think we can use `close_webinar` function here

Rados13 (approved):
There are also some problems with dependencies because some branches don't exist anymore, could you fix it?

Karolk99 (approved):

PR #159, opened 2023-02-03
Update to core 0.11

Karolk99:


Rados13 (commented):
Everything looks good but there is no branch with the name `live-dev-update-core` in membrane_rtc_engine repository.

Rados13 (changes requested):
There are some issues with dependencies.
```
** (Membrane.ParentError) Child module "Membrane.AudioFiller" is neither Membrane Element nor Bin.
Make sure that given module is the right one, implements proper behaviour
and all needed dependencies are properly specified in the Mixfile.
```

Karolk99:
There was a slight change in rtc engine with the return value when rtc engine was spawned which caused errors.

Rados13 (approved):

PR #161, opened 2023-03-01
Stateful file storage

shuntrho:


lib/membrane_live/hls/file_storage.ex
```diff
@@ -57,49 +60,63 @@ defmodule MembraneLive.HLS.FileStorage do
...
+    state = maybe_send_first_segment_notification(state, contents)
 
     {segment, partial} = Helpers.get_last_partial(contents)
+    name_without_extension = String.replace(name, ".m3u8", "")
 
     PubSub.broadcast(
       MembraneLive.PubSub,
       name_without_extension,
       {:manifest_update_partial, segment, partial}
     )
```
Karolk99, line 104: theoretically, we don't have to send partials till param `first_segment_ready` is true.

lib/membrane_live/hls/file_storage.ex
```diff
@@ -57,49 +60,63 @@ defmodule MembraneLive.HLS.FileStorage do
...
+      else
+        state
+      end
 
-    result
+    {result, state}
   end
 
   @impl true
-  def remove(_parent_id, name, _ctx, %__MODULE__{directory: location}) do
```
Karolk99, line 79: I think that state should still be the type of `__MODULE__` and config should be changed to some other type, but I'm not sure if I'm correct.

Karolk99 (changes requested):

lib/membrane_live/hls/file_storage.ex
shuntrho: I wanted to make it analogous to how it works inside Membrane's element.
For example in `RTC.Engine.Endpoint.HLS` the `Endpoint.HLS` struct is used for passing configuration to the `handle_init/2`function.
Inside `handle_init` we add some other fields to this map, but it is no longer a struct.

Karolk99: it can stay this way. But I don't think membrane elements are the best example of the general rule, because it is framework specific more than elixir specific. You can look for an example at [track module](https://github.com/membraneframework/membrane_http_adaptive_stream_plugin/blob/master/lib/membrane_http_adaptive_stream/manifest/track.ex).

Karolk99 (dismissed):

lib/membrane_live/hls/file_storage.ex
Karolk99: don't forget about it.

Karolk99 (dismissed):

Karolk99 (approved):

PR #163, opened 2023-03-06
Add measuring hls latency

Karolk99:
I don't know if we want to display it somewhere, so now it's only in the form of a console log. if you have any ideas about what exactly we can do with this statistic let me know.

### DIFF
--- a/assets/js/components/event/HlsPlayer.tsx
+++ b/assets/js/components/event/HlsPlayer.tsx
@@ -1,4 +1,4 @@
-import { useContext, useEffect, useRef } from "react";
+import { useContext, useEffect, useRef, useState } from "react";
 import HeartAnimation from "./animations/HeartAnimation";
 import ConfettiAnimation from "./animations/ConfettiAnimation";
 import HlsControlBar from "./HlsControlBar";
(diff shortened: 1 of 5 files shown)


assets/js/components/event/HlsPlayer.tsx
```diff
@@ -42,6 +43,12 @@ const HlsPlayer = ({ attachVideo, addMessage, presenterName, eventChannel, setCa
     };
   }, [addMessage]);
 
+  useEffect(() => {
+    if (!playerRef.current) return;
+    const interval = setInterval(() => doOCR(playerRef), 1000);
+    return () => clearInterval(interval);
+  }, [presenterName, attachVideo]);
```
shuntrho, line 50: We could change whether the latency is measured in this way:

```
   const [measureHlsLatency, setMeasureHlsLatency] = useState<Boolean>(false);
  (window as any).toggleHlsLatency = () => setMeasureHlsLatency((prev) => !prev);
```

We add a function to a global scope, which is probably not very elegant, however it's simple and it works.


```suggestion
  useEffect(() => {
    if (!playerRef.current || !measureHlsLatency) return;
    const interval = setInterval(() => doOCR(playerRef), 1000);
    return () => clearInterval(interval);
  }, [presenterName, attachVideo, measureHlsLatency]);
```

This way we can start measuring latency by simply calling `toggleHlsLatency()` in the console.

LVala (approved):

PR #169, opened 2023-03-24
Add private events

Karolk99:
[ML-272](https://membraneframework.atlassian.net/browse/ML-272)

[ML-272]: https://membraneframework.atlassian.net/browse/ML-272?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/membrane_live/webinars.ex
+++ b/lib/membrane_live/webinars.ex
@@ -9,7 +9,7 @@ defmodule MembraneLive.Webinars do
 
   alias MembraneLive.Webinars.Webinar
 
-  @spec list_webinars(any, boolean()) :: list(Webinar.t())
+  @spec list_webinars(binary() | :unauthorized, boolean()) :: list(Webinar.t())
   def list_webinars(_user_id, is_finished? \\ false)
 
   def list_webinars(:unauthorized, is_finished?) do


LVala (approved):
SGTM, but the solution with not logged in user is not the most elegant one (as we established), maybe somebody else has better idea?

LVala (dismissed):

lib/membrane_live_web/router.ex
```diff
@@ -26,25 +30,37 @@ defmodule MembraneLiveWeb.Router do
...
 
-    get("/", RecordingsController, :index)
     get("/:uuid", RecordingsController, :show)
   end
 
+  scope "/resources/recordings", MembraneLiveWeb do
+    pipe_through(:browser)
+    pipe_through(:auth_unrestricted)
+
+    get("/", RecordingsController, :index)
```
Rados13, line 61: So now we can list all recordings, but we can watch recording only if we are authenticated?

Karolk99: You can watch whatever you want but you have to have a link.
`:auth_unrestricted` means that it will let through all users (event not authenticated).

Rados13 (approved):

LVala (approved):

PR #171, opened 2023-04-03
Fix/dead hls endpoint

blazpie:
After we receive a `EndpointCrashed` message about HLS endpoint we're setting hls info in event state to initial values and recreating HLS endpoint

### DIFF
--- a/lib/membrane_live/event.ex
+++ b/lib/membrane_live/event.ex
@@ -83,28 +83,7 @@ defmodule MembraneLive.Event do
     Process.monitor(pid)
 
     target_segment_duration = Time.seconds(5)
-
-    endpoint = %HLS{
-      rtc_engine: pid,
-      owner: self(),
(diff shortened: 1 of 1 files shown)


lib/membrane_live/event.ex
```diff
@@ -202,6 +186,34 @@ defmodule MembraneLive.Event do
     {:noreply, state}
   end
 
+  @impl true
+  def handle_info(%Message.EndpointCrashed{endpoint_id: "hls_output"}, state) do
+    Membrane.Logger.error("HLS endpoint has crashed!")
```
Karolk99, line 191: in this case I would also log information about restarting hls_endpoint

Karolk99 (approved):

lib/membrane_live/event.ex
blazpie: Log about recreating HLS endpoint is in other callback.

PR #180, opened 2023-04-22
Add screensharing field to track metadata

Karolk99:


assets/js/utils/rtcUtils.ts
```diff
@@ -182,8 +175,8 @@ export const shareScreen = async (
 
     const newTrack = findTrackByType(client, "video", peersState);
     if (!newTrack) return peersState;
-    if (peersState.sourceIds["video"]) webrtc.replaceTrack(peersState.sourceIds["video"], newTrack);
-    else peersState.sourceIds["video"] = webrtc.addTrack(newTrack, peersState.peers[client.email].stream);
+    if (peersState.sourceIds["video"]) webrtcReplaceTrack(webrtc, newTrack, peersState, "video");
+    else peersState.sourceIds["video"] = webrtcAddTrack(webrtc, newTrack, peersState, client);
```
Rados13, line 179: This code looks very similar to the code changed below, maybe we should extract it to a function?

Rados13 (dismissed):

assets/js/utils/rtcUtils.ts
```diff
@@ -173,10 +173,7 @@ export const shareScreen = async (
 
     if (!webrtc) return peersState; // If webrtc is null, we only add media source locally
 
-    const newTrack = findTrackByType(client, "video", peersState);
-    if (!newTrack) return peersState;
```
Rados13, line 184: Doesn't it break anything? Because previously if it wasn't a new track we return from this scope, currently we only update peerState.

Rados13 (dismissed):

Rados13 (approved):

PR #191, opened 2023-07-27
add event-controller

Karolk99:
This PR is part of the migration of `MembraneLive` to `Jellyfish`.

To better understand my motivation behind this PR:
There were two modules responsible for keeping track of a webinar state: `MembraneLive.Event` and `MebraneLiveWeb.EventChannel`. 

* MembraneLive.Event - responsible for spawning RTC Engine, connecting peers, and adding HLS Endpoint. It keeps track of peers, hls and webrtc-related stuff.
* EventChannel - using `Phoenix.Presence` stores all information about all the users connected to the server.

Logic connected to maintaining webinars is partly in `EventChannel` which doesn't have its own state (only `Presence` which keeps the state of each user). The other part is inside `Event` which doesn't have the full state of the event because it's only responsible for the stream, not the whole event.

That's why I renamed `Event` to `Room` because it's only responsible for a stream, not the whole Event.
I also created a new module named `RoomController` which is responsible for controlling live times of events.
This way all the logic is in one place which is way better and easier to maintain.

Maybe in the future, It would be better to create one module responsible for both streaming and controlling event but I don't think it should be a concern of this PR.

This PR will also make it easier to rewrite `Room` to Jellyfish.


lib/membrane_live_web/channels/event_channel.ex
```diff
@@ -618,7 +558,9 @@ defmodule MembraneLiveWeb.EventChannel do
         {:error, :webinar_finished}
 
       socket.assigns.is_moderator ->
-        Event.start(socket.assigns.event_id, name: {:global, socket.assigns.event_id})
+        result = Room.start(socket.assigns.event_id, name: {:global, socket.assigns.event_id})
+        EventController.room_created(socket.assigns.event_id)
+        result
```
LVala, line 563: I didn't put much thought to it, but maybe the `EventController.room_created` should be responsible for calling `Room.start` (and change name to something appropriate)? It seems like its the only function from `Room` called in `event_channel.ex`, so maybe `EventController should be kind of a proxy between `EventChannel` and `Room`?

lib/membrane_live/room.ex
```diff
@@ -303,44 +301,24 @@ defmodule MembraneLive.Event do
...
+  @impl true
+  def handle_info({:room_controller, :kill}, state) do
+    result = Engine.terminate(state.rtc_engine, timeout: 10_000, force?: true)
 
-      {:timeout, _timer} ->
-        close_webinar(state)
+    if result == {:error, :timeout} do
+      Membrane.Logger.warn(
+        "[Event: #{state.event_id}] RTC Engine was forced kill. This can cause some problems with playing HLS playlist."
+      )
```
LVala, line 311: Let's wrap this in an API function like `Room.kill(pid)`.

lib/membrane_live/event_controller.ex
```diff
@@ -0,0 +1,198 @@
...
+      timeout: @response_timeout
+    })
+
+    timer_ref = kill_action(event_id)
+    {:noreply, put_in(state, [:events, event_id, :timer_ref], timer_ref)}
+  end
+
+  @impl true
+  def handle_info({:response, event_id, :leave}, state) do
+    send(self(), {:kill, event_id})
```
LVala, line 140:
```suggestion
    EventController.send_kill(event_id)
```
or at least be consistent and use the same target as in line 117, either `EventController` or `self()`.

LVala (changes requested):

lib/membrane_live/event_service.ex
```diff
@@ -0,0 +1,208 @@
...
+
+    {:noreply, put_in(state, [:events, event_id, :timer_ref], timer_ref)}
+  end
+
+  @impl true
+  def handle_info({:kill, event_id}, state) do
+    pid = :global.whereis_name(event_id)
+
+    {event, state} = pop_in(state, [:events, event_id])
+    {_event_id, state} = pop_in(state, [:pid_to_id, pid])
```
Rados13, line 174: I think we could use [`delete_in` here](https://hexdocs.pm/bunch/Bunch.Access.html#delete_in/2).
```suggestion
    state = delete_in(state, [:pid_to_id, pid])
```

lib/membrane_live/event_service.ex
```diff
@@ -0,0 +1,208 @@
...
+      MembraneLive.Webinars.mark_webinar_as_finished(event_id)
+    end
+
+    {:noreply, state}
+  end
+
+  @impl true
+  def handle_info({:DOWN, _ref, :process, pid, _reason}, state) do
+    {event_id, state} = pop_in(state, [:pid_to_id, pid])
+    {_event, state} = pop_in(state, [:events, event_id])
```
Rados13, line 191: Same as above.

lib/membrane_live/event_service.ex
```diff
@@ -0,0 +1,208 @@
...
+
+    unless is_nil(event_id) do
+      MembraneLiveWeb.Endpoint.broadcast!("event:" <> event_id, "finish_event", %{})
+      MembraneLive.Webinars.mark_webinar_as_finished(event_id)
+    end
```
Rados13, line 196: This part is the same as in the handler above. Maybe we could move it to a function?

lib/membrane_live/room.ex
```diff
@@ -303,44 +306,24 @@ defmodule MembraneLive.Event do
...
 
-  def handle_info({:timer_action, action}, %{timer: timer} = state) do
-    case Timer.handle_action(timer, action,
-           timeout: MembraneLive.get_env!(:last_peer_timeout_long_ms)
-         ) do
-      {:ok, timer} ->
-        {:noreply, %{state | timer: timer}}
+  @impl true
+  def handle_info(:close_room, state) do
+    result = Engine.terminate(state.rtc_engine, timeout: 10_000, force?: true)
```
Rados13, line 311: Could we move this number to the module attribute?

lib/membrane_live/room.ex
```diff
@@ -303,44 +306,24 @@ defmodule MembraneLive.Event do
...
 
-      {:timeout, _timer} ->
-        close_webinar(state)
+    if result == {:error, :timeout} do
+      Membrane.Logger.warn(
+        "[Event: #{state.event_id}] RTC Engine was forced kill. This can cause some problems with playing HLS playlist."
```
Rados13, line 315: Do we know precisely what the HLS playlist issues are?

lib/membrane_live/event_service.ex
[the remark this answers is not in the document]
Rados13: Also, all these functions as the first argument should have `event_id` and as a second `event_action`. This way, they would have the same order of arguments as `send` and `Genserver.cast`.

test/membrane_live/event_service_test.exs
Rados13, line 1: Nice tests. We could wrap all these calls like `assert_broadcast("last_viewer_active", %{}, 1_000)` to helper function `assert_last_viewer_active`, or move all these constants to a module attribute. But it is only a suggestion.

lib/membrane_live/event_service.ex
```diff
@@ -0,0 +1,208 @@
...
+  def init(_opts), do: {:ok, %__MODULE__{}}
+
+  @doc """
+  Updates event state and start timer with action (`:notify`, `:kill`) if needed.
+  Should be called every time someone joins or leaves event.
+  """
+  @spec update_event(event_action(), event_id()) :: {:event, event_action(), event_id()}
+  def update_event(action, event_id) do
+    send(EventService, {:event, action, event_id})
+  end
```
mickel8, line 56: I would split this into as many functions as there are actions

lib/membrane_live/event_service.ex
```diff
@@ -0,0 +1,208 @@
+defmodule MembraneLive.EventService do
+  @moduledoc """
+  Module responsible for controlling live times of events.
```
mickel8, line 3:
```suggestion
  Module responsible for controlling life times of events.
```

lib/membrane_live/event_service.ex
```diff
@@ -0,0 +1,208 @@
+defmodule MembraneLive.EventService do
```
mickel8, line 1: My concern is that we have one process for handling every join/leave action for every possible event which might be a bottle-neck in the future. What comes to my mind is to create process per event and rename this module to `Event`? Any thoughts?

lib/membrane_live/event_service.ex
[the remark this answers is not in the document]
Karolk99: `:kill` and `:notify` are also send using `send_after` so if I would use  `cast` instead of `send` i would have to handle to functions instead of one.

lib/membrane_live/room.ex
Karolk99: there is the `@timeout` parameter but I forgot to change this.

lib/membrane_live/event_service.ex
Karolk99: As I said in the main PR comment, I think this is a great idea but I don't want to do it in this PR.

lib/membrane_live/event_service.ex
Karolk99: I would have to use `Bunch.Access.delete_in/2` and I don't think it's worth it. What do you think?

Rados13: It was only a suggestion. I think `delete_in` is more suitable here, but no problem if you don't want to use it.

lib/membrane_live/room.ex
Karolk99: Let's say that we use ll-hls if we force kill hls endpoint recorded playlist will have partials at the end which is not good and potentially can cause some problem as it is vod (not live) playlist.

lib/membrane_live/event_service.ex
```diff
@@ -0,0 +1,229 @@
...
+    {:noreply, state}
+  end
+
+  defp notify_action(event_id),
+    do: Process.send_after(EventService, {:notify, event_id}, @notify_after)
+
+  defp kill_action(event_id), do: Process.send_after(EventService, {:kill, event_id}, @kill_after)
+
+  defp cancel_action(nil), do: nil
+  defp cancel_action(ref), do: Process.cancel_timer(ref)
```
mickel8, line 223: `start_timer` `cancel_timer` would be more descriptive

mickel8 (approved):

Rados13 (approved):

mickel8 (approved):

lib/membrane_live/room.ex
```diff
@@ -99,12 +99,16 @@ defmodule MembraneLive.Event do
...
        target_segment_duration: Time.as_milliseconds(target_segment_duration),
        start_time: nil
      }}
   end
 
+  @spec kill(pid()) :: :close_room
+  def kill(pid) do
+    send(pid, :close_room)
+  end
+
```
LVala, line 111: Could this be a `GenServer.cast`?

LVala (approved):

PR #192, opened 2023-08-01
Room refactor

Karolk99:


lib/membrane_live/chats.ex
```diff
@@ -27,19 +53,60 @@ defmodule MembraneLive.Chats do
...
-  def add_chat_message(event_id, user_name, user_email, is_auth, content, time_offset) do
-    if is_auth do
-      add_authenticated_chat_message(event_id, user_email, content, time_offset)
+  @spec add_chat_message(event_id(), user(), String.t()) :: channel_message()
+  def add_chat_message(event_id, user, content) do
+    key =
+      if user.is_auth and user.is_presenter, do: :start_timestamps, else: :client_start_timestamps
+
+    offset =
+      case :ets.lookup(key, event_id) do
```
Rados13, line 62: Isn't it a typo here? According to [docs](https://www.erlang.org/doc/man/ets.html#lookup-2), the first parameter is a table, and the second is the key. If it isn't a typo I recommend renaming the key variable.

lib/membrane_live/chats.ex
```diff
@@ -27,19 +53,60 @@ defmodule MembraneLive.Chats do
...
+        [{_key, timestamp}] ->
+          System.monotonic_time(:millisecond) - timestamp
+
+        [] ->
+          0
+      end
```
Rados13, line 68: Could we move this to a function like `calculate_offset`?

lib/membrane_live_web/channels/event_channel.ex
```diff
@@ -498,58 +476,43 @@ defmodule MembraneLiveWeb.EventChannel do
...
-        Reason: #{inspect(reason)}
+        Room: #{inspect(socket.assigns.event_id)}.
+        Reason: webinar is finished.
         """)
 
-        {:error, %{reason: "failed to start event: #{reason}"}}
+        {:error, %{reason: "failed to start event:  webinar is finished."}}
+
+      _else ->
+        {:ok, socket}
```
Rados13, line 509: Shouldn't we check if this else is not some other error?

mickel8 (commented):
Good idea with introducing well defined user db_message and channel_message maps. I lack the necessary context but in general I think the code is easier to read and maintain now :+1:

lib/membrane_live/chats.ex
Karolk99: I don't know I copied it from `event_channele.ex` but I will look at it.

lib/membrane_live_web/channels/event_channel.ex
Karolk99: If you look at the function inside the case there are four return types:
```
:ok | :already_started | {:error, :webinar_finished} |  {:error, :non_moderator}
```
only in case of `{:error, :webinar_finished}` we want to return error.

test/membrane_live/event_service_test.exs
```diff
@@ -9,8 +9,8 @@ defmodule MembraneLive.EventServiceTest do
     event_id = Ecto.UUID.generate()
     @endpoint.subscribe("event:#{event_id}")
 
-    {:ok, room_pid} = EventService.start_room(event_id)
-
+    :ok = EventService.start_room(event_id)
+    room_pid = :global.whereis_name(event_id)
```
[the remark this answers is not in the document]
Karolk99: I thought about it, but I wrap the whole `Room` using `EventService` and I thought that I shouldn't return the `Room` pid.

lib/membrane_live/event_service.ex
```diff
@@ -44,6 +51,64 @@ defmodule MembraneLive.EventService do
...
+  @spec remove_peer(event_id(), peer_id()) :: :ok | room_error()
+  def remove_peer(event_id, peer_id) do
+    case :global.whereis_name(event_id) do
+      :undefined ->
+        room_doesnt_exist_error()
+
+      pid ->
+        Room.remove_peer(pid, peer_id)
+    end
+  end
```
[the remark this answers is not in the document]
Karolk99: because those functions have different numbers of parameters I would have to write it using Lambdas 
```
def add_peer(event_id, peer_id) do
   add_peer = fn pid -> Room.add_peer(pid, peer_id) end
   proceed_if_room_exist(add_peer, event_id)
end
```
So I would prefer to stick to the current version. Or we could also use `with`:
```
defp get_room(event_id) do
   case :global.where_is(event_id) do
      :undefined ->
        room_doesnt_exist_error()

      pid ->
        {:ok, pid}
    end
end

def add_peer(event_id, peer_id) do
   with {:ok, pid} <- get_room(event_id) do
      Room.add_peer(pid, peer_id)
   end
end
```
What do you think?

mickel8 (dismissed):

lib/membrane_live/event_service.ex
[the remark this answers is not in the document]
Rados13: Between the current version and with I prefer `with`, as about creating lambdas, I don't see a problem with using them, but if you don't want to use them here ok.

lib/membrane_live_web/channels/event_channel.ex
Rados13: Then shouldn't we change it to if? Because the case with only two matches looks pretty strange.

Rados13 (dismissed):
There are two potential changes discussed in the previous comments, but besides that, PR looks good.

LVala (approved):

PR #196, opened 2024-01-29
Membrane live on jellyfish

Karolk99:


lib/membrane_live_web/channels/event_channel.ex
```diff
@@ -86,13 +89,14 @@ defmodule MembraneLiveWeb.EventChannel do
              {:ok, is_presenter} <-
```
Rados13, line 89: Shouldn't we apply the recommendation from credo?
```suggestion
             {:ok, presenter?} <-
```

lib/membrane_live_web/channels/event_channel.ex
```diff
@@ -197,16 +169,34 @@ defmodule MembraneLiveWeb.EventChannel do
...
+      response ->
+        {:reply, {:ok, response}, socket}
+    end
+  end
+
   def handle_in(
-        "am_i_main_presenter",
-        %{"is_presenter" => is_presenter},
-        %{topic: "private:" <> topic_id} = socket
+        "last_viewer_answer",
```
Rados13, line 184: Why do we receive from js code strings in `snake_case`?

lib/membrane_live_web/channels/event_channel.ex
```diff
@@ -263,193 +309,96 @@ defmodule MembraneLiveWeb.EventChannel do
...
+    case join_event(socket) do
+      {:ok, socket, token} ->
+        {:ok, _ref} =
+          Presence.update(
+            socket,
+            email,
+            &%{
+              &1
+              | is_presenter: true,
+                ## request presenting przeniesc do proposition
```
Rados13, line 341: Do we plan to address this issue or is it the TODO for future?

config/runtime.exs
```diff
@@ -118,46 +56,7 @@ config :membrane_live, MembraneLiveWeb.Endpoint, [
...
-      processors: [
-        otel_batch_processor: %{
-          exporter: exporter
-        }
-      ]
-    )
+config :jellyfish_server_sdk,
+  server_address: "localhost:5002",
+  server_api_token: "development",
+  secure?: false
```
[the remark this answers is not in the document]
Karolk99: i forgot about it, I will use envs and maybe some default values

lib/membrane_live_web/channels/event_channel.ex
Karolk99: credo only warns about functions names

kamil-stasiak (changes requested):

lib/membrane_live_web/channels/event_channel.ex
Karolk99: I don't know why it's done like that. All messages from frontend messages are in snake case so I don't think I should refactor that in this particular PR

assets/js/components/event/ControlPanel.tsx
```diff
@@ -109,127 +99,83 @@ const SettingsModal = ({ isOpen, onClose, elements }: SettingsModalProps) => {
...
+      key={key}
+      mainText={`${device.deviceInfo?.kind} source`}
+      currentSourceName={device.deviceInfo?.label}
+      sources={device.devices || []}
+      onSelectSource={(deviceId) => {
+        device.start(deviceId);
+      }}
+    />
+  );
+};
```
kamil-stasiak, line 117: It's a component, so you could use JSX to emphasise it:

```ts
const DeviceButton = ({ key, device }: {
  key: string,
  device: UseMicrophoneResult<TrackMetadata> | UseCameraResult<TrackMetadata>
}) => <DropdownButton
  key={key}
  mainText={`${device.deviceInfo?.kind} source`}
  currentSourceName={device.deviceInfo?.label}
  sources={device.devices || []}
  onSelectSource={(deviceId) => {
    device.start(deviceId);
  }}
/>
```

and then use it like this:

```ts
<SettingsModal
              isOpen={isOpen}
              onClose={onClose}
              elements={[
                <DeviceButton key={"1"} device={microphone}/>,
                <DeviceButton key={"2"} device={camera}/>
              ]}
            />

```

assets/js/components/event/ControlPanel.tsx
```diff
@@ -109,127 +99,83 @@ const SettingsModal = ({ isOpen, onClose, elements }: SettingsModalProps) => {
...
+          <IconEnabled className="PanelButton Enabled" />
+        ) : (
+          <IconDisabled className="PanelButton Disabled" />
+        )
+      }
+      onClick={() => {
+        device.setEnable(!device.enabled);
+      }}
+    />
+  );
```
kamil-stasiak, line 137: It's also a component, you could rewrite it as `getDropdownButton`

kamil-stasiak (approved):

lib/membrane_live/room.ex
```diff
@@ -101,10 +102,10 @@ defmodule MembraneLive.Room do
 
   @impl true
   def handle_call({:remove_peer, peer_id}, _from, state) do
+    state = handle_peer_left(state, peer_id)
+
     case Jellyfish.Room.delete_peer(state.client, state.room.id, peer_id) do
       :ok ->
```
Rados13, line 108: So now we will remove the peer from the state even though it can still be on jellyfish?

lib/membrane_live/room.ex
```diff
@@ -142,7 +143,7 @@ defmodule MembraneLive.Room do
     new_state = %{state | playlist_ready?: false}
     send_broadcast(new_state)
 
-    {:noreply, state}
+    {:stop, :normal, state}
```
Rados13, line 146: Shouldn't it be a different reason when `ComponentCrashed`?

Rados13 (approved):

